PR #3, opened 2015-11-24
Allow passing in `customCSS` path option.

jgerigmeyer:


carljm:
I only vaguely understand the gulpfile changes, which seem only vaguely related to this feature :-) But otherwise :+1:

jgerigmeyer:
That's fair, since the gulpfile changes aren't really related. :)

The only notable change is L132, which fixes a bug I discovered where the tests can't be running at the same time as a ``compile``, but I'm not 100% sure why...

PR #23, opened 2016-06-10
Improved previews

mirisuzanne:
Trying to simplify the path from adding a color/font to seeing it show up in the style guide:
- [x] color styles are added to color palette inline
- [x] font styles are added to font specimen inline
- [x] add colors to sass config and json export simultaneously
- [x] add fonts to sass config and json export simultaneously
- [x] add sizes to sass config and json export simultaneously
- [x] document previews & sass utilities


mirisuzanne:
@davisagli when we have a chance (tomorrow?) I'd like to review with you what I've done, and what could use some help.

mirisuzanne:
@jgerigmeyer I added tabs in this update, but the JS isn't in place for it. That something you can add easily? Otherwise this just needs documentation, and it's ready for merge.

mirisuzanne:
note, I think we'll also want basic toggles for an off-canvas layout soon (not in this branch). If that JS is related, may as well bring the whole thing over.

jgerigmeyer:
@mirisuzanne Ok, I think that gives you what you need?

templates/item/macros.j2
```diff
@@ -0,0 +1,211 @@
...
+  {% endif %}
+{%- endmacro %}
+
+
+{# @require annotation #}
+{% macro require(item) -%}
+  {% if item.require.length > 0 %}
+    <div data-sassdoc="item-description">
+      <h5 class="item-subtitle">Requires</h5>
+      <ul>
```
stacyk: `<li data-sassdoc="requires-item">`, `<li data-sassdoc="link-item">`, ` <li data-sassdoc="see-item">` these types of list items need a left margin, or perhaps a class on the <ul> to include the margin, or something else to hook into like `main` `li` ... I guess the nav isn't using list items (should we be?) so maybe no specificity is needed?

scss/previews/_font.scss
```diff
@@ -0,0 +1,52 @@
+// Font Specimen
```
mirisuzanne: Done in 09ea105

scss/previews/_examples.scss
```diff
@@ -0,0 +1,41 @@
...
+
+[data-sassdoc-example-panel] {
+  border: 1px solid color('border');
+  max-height: size('rhythm') * 12;
+  max-width: 100%;
+  overflow: auto;
+}
+
+[data-sassdoc-example-panel='output'] {
+  padding: size('shim');
```
mirisuzanne: Yeah, this is another example of some meta-documentation confusing what you see. The output always has padding, so that you don't confuse the frame with the design inside it. In this case, we're demonstrating what a code-block looks like — so this code-block is example output, while the other code-block is part of the actual design. This will not be a common use-case.

mirisuzanne:
@stacyk I know there is a lot more to fix in terms of the overall layout and design. I pushed a few fixes based on your comments, but this branch is specifically about showing better previews, and I'd like to save any other changes for another branch — as long as the previews themselves are better.

PR #34, opened 2016-09-14
macro documentation [WIP]

mirisuzanne:
document macros directly in the j2/njk markup, using sassdoc-like [annotations](http://sassdoc.com/annotations/)

### DIFF
--- a/templates/item/preview.macros.j2
+++ b/templates/item/preview.macros.j2
@@ -1,3 +1,36 @@
+
+{### Code Block
+
+Generate a syntax-highlighted pre/code block.
+
+@access public
+@group code-blocks
+@link https://highlightjs.org/
+
+@author Miriam Suzanne
+@since 0.1.1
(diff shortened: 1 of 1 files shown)


davisagli:
How are you thinking we would tell sassdoc to scan for this? Would we reference a particular njk file or macro from within a Sass file? Or would we want herman to find all the njk files in a list of paths and create a new page in the documentation for each one that contains annotations (thus making documentation in templates more of an equal peer to documentation in sass rather than secondary)?

mirisuzanne:
My thought was that we pass a nunjucks folder to scan, similar to the Sass — and add the documentation to `@group` pages in a similar way (not based on file or path). Basically, the equal peer option.

davisagli:
Cool, that's my preference too in terms of functionality. I'll have to look at whether we can implement it without changing sassdoc itself.

PR #35, opened 2016-09-29
Mark `minifiedIcons` file as `|safe`.

jgerigmeyer:


templates/base.j2
```diff
@@ -14,7 +14,7 @@
 </head>
 <body>
 
-  {{ iconsSvg }}
```
jgerigmeyer: So... is this a terrible idea from a security standpoint? It's required to make this feature work with recent versions of Nunjucks.

davisagli:
Where does `iconsSvg` come from and do we trust it not to include malicious markup?

jgerigmeyer:
We accept a filename arg, and then read the file and pass it in directly. So... no?

https://github.com/oddbird/sassdoc-theme-herman/blob/master/index.js#L214-L219

jgerigmeyer:
Basically, it requires trusting that the file at that path, on any machine where `webpack` is run for this project, contains valid html and not malicious markup.

davisagli:
Well, we're trusting the person who configures herman to use the file they intend to. That seems reasonable.

templates/base.j2
wlonk: So, if this was working in an older version of Nunjucks, I expect it would have exactly the same security implications—it would have worked by implicitly treating this data as safe. So, if it hasn't been an issue, maybe it's not a problem? This only exposes things that are under the control of the person building the docs, yeah?

jgerigmeyer:
I'm convinced. Thanks!

PR #40, opened 2016-12-13
Add font-variant override "show" argument

mirisuzanne:


davisagli:
Looks good as long as you mention it in the changelog

mirisuzanne:
Thanks! Adding a bit more documentation inline as well...

PR #41, opened 2016-12-15
Fix see links absolute->relative

mirisuzanne:


mirisuzanne:
@davisagli - Sassdoc defaults to a single page. We added the feature that breaks groups into pages. At some point I'd like to consider nested grouping, but that would be a new issue. It's not supported yet by SassDoc or Herman.

PR #43, opened 2016-12-15
Add customHead option

### LINKED ISSUE #42: I want to pass in arbitrary html to the document <head>
Especially for use with font CDN's like Typekit, that require an injection script something like:

```
<script src="https://use.typekit.net/unique.js"></script>
<script>try{Typekit.load({ async: true });}catch(e){}</script>
```

davisagli:
Fixes #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Releases
 
 - Change absolute `@see` links to relative paths,
   so we're not assuming a root URL for the styleguide.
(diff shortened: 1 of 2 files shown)


davisagli:
To use this you pass the markup to inject as the `customHead` option when invoking sassdoc (the same place you pass `customCSS`)

PR #46, opened 2017-01-13
render HTML examples

### LINKED ISSUE #45: I want to see output from HTML examples
Right now you can create `njk` examples, that will display with both code and output:

```scss
/// @example njk
///   {% import 'question/macros.njk' as question %}
///   {{ question.circle() }}
///   {{ question.circle(scope='answer', is_accepted=true) }}
///   {{ question.circle(scope='answer', is_best=true) }}
```

njk
output

Sometimes you just need simple html examples that work the same way. For now, we've been marking those as `njk` on projects that allow it — but when we have a project not using nunjucks, it would still be nice to display plain html examples. 

```scss
/// @example html
///   <button>example button</button
```

davisagli:
fixes #45


README.md
```diff
@@ -102,6 +102,7 @@ Releases
 
 - Change absolute `@see` links to relative paths,
   so we're not assuming a root URL for the styleguide.
+- Render type `html` examples to HTML.
```
mirisuzanne: It looks like 0.5.3 has already been released, so this should all be updated to match the missing changes.

mirisuzanne:
Works great! I added an example.

davisagli:
I rebased on up-to-date master to fix the changelog.

PR #71, opened 2017-05-19
Render prose documentation comments not attached to Sass blocks

davisagli:
This makes it possible to add narrative prose that isn't attached to a particular Sass block.

Note: depends on an upstream feature in sassdoc: https://github.com/SassDoc/sassdoc/pull/498


davisagli:
This is no longer blocked; we just need to update the branch, upgrade sassdoc and do review.

mirisuzanne:
I don't see a new release of sassdoc. Are we still blocked by that?

davisagli:
@mirisuzanne sassdoc 2.3.0 has the necessary change

index.js
```diff
@@ -74,6 +74,27 @@ var prepareContext = function(ctx) {
...
+    if (item.context.type === 'unknown') {
+      item.context.type = 'prose';
+      item.context.line.end = item.context.line.start;
+    }
+    if (item.context.line.start > item.commentRange.end + 1) {
+      item.context = {
+        type: 'prose',
+        line: item.commentRange
+      };
+    }
```
jgerigmeyer: This block breaks cases where the Sass block selector is more than one line, which causes a line between `commentRange.end` and `context.line.start`. E.g. `_highlight.scss` `Code Blocks` section:

```scss
// Code Blocks
/// @name Code Blocks
/// @group code
/// @access private
/// @requires {function} hljs-gradient
[data-sassdoc='hljs'],
[data-sassdoc-page='index'] pre {
  align-items: stretch;
  background-color: color('hljs-background');
  color: color('hljs-text');
  display: flex;
  position: relative;
  ...
```

jgerigmeyer:
There's still one bug here (hoping @davisagli can dig into it or provide some context), but @mirisuzanne I think you have what you need to at least see what prose comments look like and do any template/style cleanup you'd like.

index.js
davisagli: Hmm... The case I'm trying to detect here is where there are blank lines in between a sassdoc comment and a selector (because in that case the comment is probably meant to be prose, but the scss-comment-parser associates it with the selector). Maybe we can actually get the contents of line number `commentRange.end + 1` and see if it's blank.

jgerigmeyer: Makes sense. Is that something you have time to look at this week?

davisagli: Yeah, I can.

mirisuzanne:
Prose blocks are mixed in with everything, based on alphabetical order of the contents. That's probably not great for most use-cases. There is an easy work-around, using the `@name` annotation, which is hidden on prose blocks, but still controls the order. Is that too hacky to point out in the docs? Do we need a better way to handle ordering?

davisagli:
@mirisuzanne oh, I was assuming documentation gets rendered in the same order that things appear in the source code. Are we intentionally alphabetizing or is it a side effect of something? Alphabetizing doesn't seem like a good idea to me, at least if it's mandatory.

jgerigmeyer:
@mirisuzanne I removed the `@name` lines, and ordering of prose blocks is consistent for me, as it falls in the scss file. I'm not seeing anything alphabetized.

mirisuzanne:
@davisagli @jgerigmeyer oh, odd. Now I see what you're seeing - no alphabetizing. It was for sure out of order for me yesterday. Oh well. If it happens again, we can try and track it down. ¯\_(ツ)_/¯

Issue #65, opened 2017-05-19
Show sass & css, nunjucks & html beside one another


stacyk:
![screen shot 2017-05-19 at 11 13 12 am](https://cloud.githubusercontent.com/assets/1581694/26263058/f72bd808-3c8b-11e7-9ac9-e6aa29580a81.jpg)

Not necessarily the design I am suggesting, but I think it is helpful to see side by side :)

stacyk:
This would be a tabbed set that includes 2 top level  "Styles" and "Markup"
Styles would include 1-2 panels: `SCSS` and `CSS` if available.
Markup would include 2 panels: `NJK' and 'HTML`

jgerigmeyer:
Closed by #72.

PR #72, opened 2017-05-20
Example panels

stacyk:
Related to issuse #65 #46 #45


mirisuzanne:
This is not highly styled, but the styles are changing soon anyway.

scss/previews/_examples.scss
```diff
@@ -23,19 +15,44 @@
...
+  padding: 0 size('shim');
+
+  + [data-sassdoc] {
+    flex: 1 1 auto;
+  }
 }
 
-[data-sassdoc-example-panel='output'] {
+[data-sassdoc-rendered] {
   padding: size('shim');
```
stacyk: This makes the rendered panel more narrow than the tab groups. We can eliminate l/r padding. Maybe in favor of using t/b margins instead?

templates/item/macros.j2
```diff
@@ -168,58 +168,84 @@ ${{ item.context.name }}: {{ item.context.value }}{% if item.context.scope != 'p
...
-          class='input'
-          ) }}
+          active=true if loop.first else false
+          ) %}
+
+          <div data-sassdoc-example="{{ example.type }}">
+            <div data-sassdoc-title="example">
+              {{ example.type }}
+              {% if example.description %}
+                – <em>{{ example.description|striptags }}</em>
```
stacyk: I wonder if we even want any italic in order to avoid loading another font file. Unless we stick to system fonts only, which is something we sorta discussed yesterday. Maybe a span with a class so we can style it differently might work?

templates/item/macros.j2
```diff
@@ -168,58 +168,84 @@ ${{ item.context.name }}: {{ item.context.value }}{% if item.context.scope != 'p
...
+              {% endif %}
+            </div>
+            {{ code_block(language=example.type, content=example.code|escape|safe) }}
+          </div>
+
+          {% if example.rendered %}
+            {% set render_type = render_types[example.type] %}
+            <div data-sassdoc-example="{{ render_type }}">
+              <div data-sassdoc-title="example">
+                {{ render_type }} – <em>compiled</em>
```
stacyk: Same comment as above with the `<em>`

templates/item/macros.j2
```diff
@@ -168,58 +168,84 @@ ${{ item.context.name }}: {{ item.context.value }}{% if item.context.scope != 'p
...
-          <span data-sassdoc="example-tab-solo">{{ example.type }}</span>
-        </div>
-        <div data-sassdoc-example-panel="{{ example.type }}">
-          {{ code|safe }}
-        </div>
+          {% set next = examples_list[loop.index] %}
+          {% if (next.description|striptags == 'compiled') %}
+            <div data-sassdoc-example="{{ next.type }}">
+              <div data-sassdoc-title="example">
+                {{ next.type }} – <em>compiled</em>
```
stacyk: Same comment as above with the `<em>`

Issue #37, opened 2016-11-22
I want Sass examples to be compiled automatically


mirisuzanne:
Same as `njk` examples.

davisagli:
Is this for documenting how to use Sass macros and functions? So we would need to render the Sass example in an environment that includes the Sass sassdoc is being run on, right?

mirisuzanne:
That's right, yes.

davisagli:
l'm not quite sure how to tackle this. The challenges are:
1. sassdoc operates on a list of sass files, and there's no info about which one is meant to be the main entry point so that we know what to import as the environment to render the example in. So I don't think there's a way around adding a config option to specify this.
2. Even if we know what to import, I don't know a way to make node-sass operate in two stages so that it first imports all the Sass without rendering it, and then renders just the example using the former as an environment.

PR #75, opened 2017-05-20
Compile Sass examples automatically

wlonk:
See #37.


wlonk:
Blech, tests passed locally.

PR #78, opened 2017-05-20
Nest Herman opts

jgerigmeyer:


gulpfile.js
```diff
@@ -157,10 +157,12 @@ gulp.task('compile', [ 'sass', 'minify' ], function () {
...
     theme: './',
-    sassjsonfile: './dist/css/json.css',
-    templatepath: path.join(__dirname, 'templates'),
-    sassincludepaths: [path.join(__dirname, 'scss')],
-    sassincludes: [ 'utilities', 'config/manifest' ],
+    herman: {
+      sassjsonfile: './dist/css/json.css',
+      templatepath: path.join(__dirname, 'templates'),
+      sassincludepaths: [path.join(__dirname, 'scss')],
+      sassincludes: [ 'utilities', 'config/manifest' ]
```
wlonk: We could even make all the sass things be in their own namespace?

wlonk: Also, the arrays on these two lines have inconsistent spacing around them.

index.js
```diff
@@ -439,10 +440,10 @@ module.exports.annotations = [
...
               exampleItem.rendered = undefined;
               try {
-                if (env.sassincludes) {
-                  for (var i = env.sassincludes.length - 1; i >= 0; i = i - 1) {
-                    sassData =
-                      "@import '" + env.sassincludes[i] + "';\n" + sassData;
+                if (env.herman.sassincludes) {
+                  var arr = env.herman.sassincludes;
+                  for (var i = arr.length - 1; i >= 0; i = i - 1) {
+                    sassData = "@import '" + arr[i] + "';\n" + sassData;
```
wlonk: Can this be `for ... of` now?

gulpfile.js
jgerigmeyer: According to our current linter rules, arrays with one item have no spaces (`['foo']`) while arrays with multiple items have spaces (`[ 'foo', 'bar' ]`). ¯\_(ツ)_/¯

index.js
jgerigmeyer: Hm, I guess so? Our linter currently balks at any ES6 to be safe, even though _some_ ES6 features are supported by Node v4.

The other thing here is that I want to iterate over the array in reverse, but I don't want to `.reverse()` the array every time. So if switching to a `for...of` loop, I'd need to first `.reverse()` the array just once, and then iterate for every `item`.

gulpfile.js
jgerigmeyer: Addressed in #84.

PR #86, opened 2017-05-21
Use .sassdocrc for subproject config.

### LINKED ISSUE #85: Subprojects should use .sassdocrc config.

jgerigmeyer:
Fixes #85.

### DIFF
--- a/index.js
+++ b/index.js
@@ -203,7 +203,8 @@ var parseSubprojects = function (ctx) {
         // Load .sassdocrc configuration from subproject directory
         config = yaml.safeLoad(fs.readFileSync(configFile, 'utf-8'));
       } catch (err) {
-        ctx.logger.warn('No .sassdocrc found for subproject: ' + name);
+        ctx.logger.warn(
+          'Invalid or no .sassdocrc found for subproject: ' + name);
       }
       if (!config.description && !config.descriptionPath) {
         // Set default descriptionPath for subproject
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -194,9 +195,34 @@ var parseSubprojects = function (ctx) {
...
-      var promise = sassdoc.parse(prjCtx.src, prjCtx).then(function (data) {
+    ctx.herman.subprojects.forEach(function (name) {
+      var prjPath = './node_modules/' + name + '/';
+      var configFile = prjPath + '.sassdocrc';
+      var config = {};
+      try {
+        // Load .sassdocrc configuration from subproject directory
+        config = yaml.safeLoad(fs.readFileSync(configFile, 'utf-8'));
+      } catch (err) {
+        ctx.logger.warn('No .sassdocrc found for subproject: ' + name);
```
davisagli: This message is misleading if the problem is malformatted YAML rather than a missing file.

index.js
```diff
@@ -194,9 +195,34 @@ var parseSubprojects = function (ctx) {
...
+      }
+      if (!config.description && !config.descriptionPath) {
+        // Set default descriptionPath for subproject
+        config.descriptionPath = prjPath + 'README.md';
+      }
+      var src;
+      if (config.src) {
+        // Set subproject src files based on .sassdocrc `src` option
+        src = prjPath + config.src;
```
davisagli: Should only do this if the existing config.src is a relative path.

PR #98, opened 2017-05-22
[style-bugs] smaller default title, tag-style, smaller icon, more main padding [style-bugs]

stacyk:
#90

Issue #93, opened 2017-05-22
How shall we organize all these goram annotations?


mirisuzanne:
Top Priority:
- [x] Type & Name & Variable-Type (e.g `@mixin section-arrow()` or `$colors (Map)`)
- [x] Alias (alias-for, or aliased as)
- [x] Source (shows the actual code)
- [x] Description (short markdown description)
- [x] Parameters or Properties - currently gross tables

Related to Parameters and Properties
- [x] Content
- [x] Output or Return
- [x] Throw

Lists (mostly just links):
- [x] Link
- [x] See
- [x] Require
- [x] Todo (may have a bit more description)

Examples:
- [x] Tabbed code-blocks
- [x] Rendered examples

Herman Previews:
- [ ] Color Palette (already has a story)
- [ ] Font Specimen
- [ ] Icons

Metadata:
- [x] Author
- [x] Access (public or private)
- [x] Deprecated
- [x] Since

[SassDoc Annotations](http://sassdoc.com/annotations/) and [Demo Gallery](http://sassdoc.com/theme-gallery/)

mirisuzanne:
I think this is complete, with detailed stories for any remaining style/functionality details.

Issue #77, opened 2017-05-20
Create sample sass file for documentation


mirisuzanne:
We need a way to document all the possible annotations, without having to use them for the actual Herman styleguide. 

PR #83

PR #99, opened 2017-05-22
[style-cleanup-mia] Style cleanup mia

### LINKED ISSUE #5: Render SassDoc `@alias`

### LINKED ISSUE #49: Need to remove margin from <p> in parameters table
Most, if not all, of our css has a margin on the `<p>` element. Sassdoc parameters table description automatically wraps that content in a `<p>`. Removing the top/bottom margin from this sassdoc param table `<p>` would allow the other items in the row to line up properly.  

OR, as we spoke about in Slack, maybe we just need to do this:

Adding `|striptags` instead of `|safe` works - just means we also can't include links or emphasis.

### LINKED ISSUE #53: Don't list identical requirements
Example from True:

```
Requires
- [variable] _true-test-result
- [variable] _true-test-result
- [variable] _true-test-result
- [variable] _true-test-result
```

mirisuzanne:
- Add `@alias` styles [fixes #5]
- Filter identical requirements [fixes #53] 
- Properly handle `<p>` tags on descriptions [fixes #49]
- Add initial samples for testing [related to #77] 
- Organize existing annotations and annotation-macros [related to #93]
- Style parameter lists without that nasty table [related to #93]
- Project-links only show on index pages [needs work for subprojects in #100]
- Add extra padding on large screens [likely conflict with #98 - oops]
- Fix bug in `attr_if` macro
- move `code_block` into utilities
- Fix `content`/`caller()` bugs in utilities
- Add group configuration where most needed


scss/samples/_variables.scss
```diff
@@ -0,0 +1,40 @@
...
+/// The basics for documenting your average non-map variable.
+/// @group samples
+/// @type *
+/// @requires $private
+$basic-value: if(not 'social', $private, 'hello world');
+
+
+// Map Value
+// ---------
+/// The basics for documenting your average non-map variable.
```
davisagli: s/non-map/map/

mirisuzanne:
@david yeah - I've been changing over to classes as I get to it. Maybe that should become a concerted effort at some point.

scss/component/_item.scss
```diff
@@ -1,39 +1,70 @@
...
+.alias-title {
+  font-style: normal;
+  font-weight: bold;
+}
+
+.param-list {
+  margin-bottom: size('shim');
+}
+
+.param-title {
```
stacyk: This feels a little tight to me. Could we add a half-shim margin-bottom or maybe it would be better to add a margin-top to the `.param-details` instead?

scss/samples/_variables.scss
```diff
@@ -0,0 +1,40 @@
...
+/// @access private
+/// @type *
+$private: 'for internal use only';
+
+
+// Basic Value
+// -----------
+/// The basics for documenting your average non-map variable.
+/// @group samples
+/// @type *
```
stacyk: `/// @type *` this confused me at first. Does it just mean it is a global variable? If it weren't, what would type be?

scss/samples/_variables.scss
```diff
@@ -0,0 +1,40 @@
...
+// ---------
+/// The basics for documenting your average non-map variable.
+/// @group samples
+/// @type map
+/// @requires $basic-value -
+///   We can require other variables, and explain why
+/// @requires $private -
+///   We can also override the link <http://oddbird.net>
+/// @property {string} 'morning' [$basic-value] -
+///   What to say in the morning when you get up
```
stacyk: What if you hate mornings?

templates/item/annotation.macros.j2
```diff
@@ -0,0 +1,297 @@
...
+    note=note,
+    description=item.context.description
+    ) }}
+{% endmacro %}
+
+
+{# @todo annotation #}
+{% macro todo(item) %}
+  {% if (item.todo.length > 0) %}
+    {% call item_section('todo', 'Todo List') %}
```
stacyk: Should the title read `To Do List`?

scss/samples/_variables.scss
mirisuzanne: Type can be things like `string`, `map`, `color`, `length`, `number`, `list`, etc. That's all in the SassDoc annotations documents.

scss/samples/_variables.scss
mirisuzanne: ha. I do! But you can say `hello world` in a grumpy voice, if you like.

templates/item/annotation.macros.j2
mirisuzanne: This sent @SondraE down a rabbit-hole of googling. A google-hole? Seems the "proper" way, according to various dictionaries, would be "to-do list". :)

davisagli: Unless it's a list of everything in Spanish

templates/base.j2
```diff
@@ -55,20 +52,9 @@
...
-            {% if package.repository.url %}
-              <li data-sassdoc="project-link-item">
-                <a href="{{ package.repository.url }}" data-sassdoc="project-link">View Source</a>
-              </li>
-            {% endif %}
-          </ul>
-        {% endif %}
+        {# if activeProject %}
+          <a href="{{ basePath }}{{ activeProject }}/index.html">{{ activeProject }}</a>
+        {% endif #}
```
jgerigmeyer: Why is this commented out?

templates/group.j2
```diff
@@ -1,8 +1,6 @@
 {% extends 'base.j2' %}
 
 {% block main %}
-  <h2 data-sassdoc="page-title">{{ pageTitle }}</h2>
```
jgerigmeyer: Hm... On smaller sizes (with the menu collapsed), this feels like useful context to me.

mirisuzanne: Ahh, both of these are because I didn't finish the breadcrumb Sondra requested in place of that mid-level header. PR now at #104

PR #114, opened 2017-05-30
[google-fonts] Use google fonts

### LINKED ISSUE #113: Can we avoid shipping font files?
Maybe use googlefonts instead?

mirisuzanne:
Fixes #113


templates/base.j2
```diff
@@ -8,6 +8,7 @@
   <title>{% block title %}{{ project_title }} Documentation{% endblock %}</title>
 
   <link href="{{ shortcutIcon.url }}" rel="shortcut icon">
+  <link href="https://fonts.googleapis.com/css?family=Source+Code+Pro|Source+Sans+Pro" rel="stylesheet">
```
stacyk: I saw that the preview for these two typefaces is requesting normal, bold, italic, bold italic ...
1. I don't think Source Code Pro has italic?
2. Wouldn't we need to declare it like this: 

`<link href="https://fonts.googleapis.com/css?family=Source+Code+Pro:400,700|Source+Sans+Pro:400,400i,700,700i" rel="stylesheet">`

mirisuzanne: Thanks! I haven't used google fonts much, so wasn't sure. I'll clean it up.

PR #127, opened 2017-06-08
Better example display, and code spacing

### LINKED ISSUE #116: Fix alias-links in `used-by`
Since alias mixins/functions often require the original mixin/function, they appear in that functions used-by, with a link, even if `display: alias` is turned off.

### LINKED ISSUE #120: Remove extra margin below `source` code block

### LINKED ISSUE #121: Add `:` after map property keys

### LINKED ISSUE #122: same-language tabs break, because IDs match
Related to #106.

mirisuzanne:
Fixes #120 
Fixes #122 
Fixes #121 
Fixes #116


mirisuzanne:
@stacyk Right now there is no way to say "these two examples are related" - which means I'm not sure how we would make the spacing/rhythm any more clear. I think it would be great to have an example-group option, but that would have to be a new issue.

PR #140, opened 2017-07-04
Account for multi-line names when detecting prose

davisagli:
This fixes the issue discussed at https://github.com/oddbird/sassdoc-theme-herman/pull/71#discussion_r124052871

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,9 +86,10 @@ var prepareContext = function(ctx) {
       item.context.type = 'prose';
       item.context.line.end = item.context.line.start;
     }
-    // @@@ This breaks cases where a Sass block selector is more than 2 lines.
-    // https://github.com/oddbird/sassdoc-theme-herman/pull/71#pullrequestreview-46309820
-    if (item.context.line.start > item.commentRange.end + 2) {
+    // Consider it to be prose if it's separated from the next Sass block
+    // by any blank lines.
+    var lineCount = item.context.name.split('\n').length;
+    if (item.context.line.start > item.commentRange.end + lineCount) {
       item.context = {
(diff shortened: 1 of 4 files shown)


jgerigmeyer:
@davisagli This still breaks in the same place. Given the following code:

```scss
// Code Blocks
/// @name Code Blocks
/// @group code
/// @access private
/// @requires {function} hljs-gradient
[data-sassdoc='hljs'],
[data-sassdoc-page='index'] pre {
  align-items: stretch;
  ...
```

I don't see an obvious way for Herman to tell the difference between a blank line and the first part of a selector: `[data-sassdoc='hljs'],`. For this section, the `name` is `Code Blocks` (1 line), but the `commentRange.end` is `/// @requires {function} hljs-gradient` while the `line.start` is `[data-sassdoc-page='index'] pre {`. So it's like a blank line in between, as far as the data shows.

davisagli:
@jgerigmeyer it should be accounting for that if context.line.start is greater than commentRange.end plus the number of lines in context.name. But I do see now that it's breaking somewhere while trying to render that group, so I'll look again.

davisagli:
oh, that case has a custom @name, doesn't it

davisagli:
Hmm...mixins whose parameters span multiple lines have a context.line.start that corresponds to the first line, not the line where the curly-block begins. So now I'm thinking there's a bug in the line numbering in the support that Carl added to scss-comment-parser to handle plain CSS selectors.

davisagli:
@jgerigmeyer okay, *now* I think it'll take multiline selectors into account properly, by preserving the original name when it's customized by the `@name` annotation.

PR #145, opened 2017-08-21
[simpler-utils] Simplified Sass Utilities

### LINKED ISSUE #123: I want a better design for color palette previews
Some ideas...
- [x] Combined palette, where one color may have multiple names
- [x] More space-conscious layout
- [x] Only show the most useful details (no rgba)

### LINKED ISSUE #143: Make sure CSS exports make sense without Accoutrement

### LINKED ISSUE #144: I want Sass tests for Herman utilities

mirisuzanne:
Fixes #143 (utils) 
Fixes #144 (tests)
Fixes #123 (grouped-colors)

- [x] Blocked by https://github.com/oddbird/accoutrement-scale/issues/11
- [x] Fix linting error
- [x] Update Changelog
- [x] Add color-conversion filter to templates


mirisuzanne:
@jgerigmeyer There's still a bug in the js linter that I can't track down. Can you take a look?

jgerigmeyer:
I think this is ready for review/merge (or styles cleanup, as needed).

jgerigmeyer:
@mirisuzanne I cleaned up and merged #146 into this branch; both are ready for cleanup/review/merge.

templates/item/preview.macros.j2
```diff
@@ -31,24 +32,35 @@
...
-        <div data-sassdoc="color-overlay" style="background-color: {{ value.output }};"></div>
-      </div>
-      <div data-sassdoc="color-info">
-        <strong data-sassdoc="color-name">{{ name }}</strong>
-        {% for type, color in value %}
-          <div data-sassdoc="color-value">
-            {{ color }}
+  {% call show.item_section('color-palette', 'Color Preview') %}
+    <div class="color-palette">
+      {% for value, items in data|list|groupby('value') %}
```
mirisuzanne: This is much nicer than my hacky loop - thanks! :)

jgerigmeyer: Agreed, and it preserves the original ordering (which is a nice side-effect, I think).

mirisuzanne:
@stacyk I think this is just waiting on your review

scss/previews/_color.scss
```diff
@@ -5,35 +5,50 @@
...
 // -------------
 
-[data-sassdoc='color-palette'] {
-  margin: size('gutter') 0;
+.color-palette {
+  display: flex;
+  flex-wrap: wrap;
+  margin: negative('shim');
+
+  @supports (display: grid) {
```
stacyk: I've been wondering about this. If a browser doesn't support grid, it will ignore all grid-* properties so the only thing you really need is the margin in this @supports statement. I know we get an error message in our current lint config though when you use the same property twice. I think this won't matter much here, but the parts when we aren't declaring a new margin (or other non-grid property) then we could clean up a number of nested queries going forward. Thoughts on that topic in general (not specific to this example)?

templates/item/preview.macros.j2
```diff
@@ -31,24 +32,35 @@
...
+            <div class="color-overlay" style="background-color: {{ value }};"></div>
           </div>
-        {% endfor %}
-      </div>
+          <div class="color-info">
+            <strong class="color-name">
+              {% for item in items %}
+                <span class="color-name-option">{{ item.key }}{% if not loop.last %},&nbsp;{% endif %}</span>
+              {% endfor %}
+            </strong>
```
stacyk: I still think it would be cool to have the css syntax here as a copyable line too: 
```
color('{{ item.key }}');
```

templates/item/preview.macros.j2
```diff
@@ -31,24 +32,35 @@
...
+            {% set color = value|colors %}
+            {% if color %}
+              {% for format in ['hex', 'rgb', 'hsl'] %}
+                {% if color[format] %}
+                  <code class="color-value">{{ color[format] }}</code>
```
stacyk: More of a style note: I wonder if the inner box shadow gets to be a little much around each of these values?

hmm, just kidding. I kinda like it compared to not using the box-shadow or background:
no-vs-yesbxshadow

scss/config/_colors.scss
```diff
@@ -22,14 +22,14 @@
...
-///   @include herman-add-colors('brand-colors', $brand-colors);
+///   @include accoutrement-add('colors', 'brand-colors', $brand-colors);
 $brand-colors: (
   'brand-orange': hsl(24, 100%, 39%),
   'brand-blue': hsl(195, 85%, 35%),
   'brand-pink': hsl(330, 85%, 48%),
 );
 
-@include herman-add-colors('brand-colors', $brand-colors);
+@include accoutrement-add('colors', 'brand-colors', $brand-colors);
```
stacyk: I wonder if the 'name' can somehow be pulled from the $mapname?

The only time we don't use the same exact naming here is for the font family names. So we use 'sans' but the map is called $font-sans. But maybe that is just our own use case?

scss/previews/_color.scss
mirisuzanne: I agree. Wherever it's easy to do without the feature-query, we just do that. Here, since we need the query for one value, I thought it was cleaner to just keep them all together. I don't feel strongly about that.

templates/item/preview.macros.j2
mirisuzanne: Yeah, that's a cool idea – and I can imagine how it would work. But it sounds like a new feature to me, with a setting of it's own. Want to file an issue for that?

templates/item/preview.macros.j2
mirisuzanne: I'd go either way on this. Some advantages to removing that extra padding as well… @SondraE, want to weigh in on your favorite?

scss/config/_colors.scss
mirisuzanne: I agree it would be nice, but I can't figure out any way to do that. Sass isn't able to get the name of a variable - only the value - and we need both Sass and Herman to agree on the name.

This isn't really about changing the name or keeping it the same, it's just about telling Sass what the name is, so that Herman can find it. Which is ugly, but I don't have a better idea.

PR #165, opened 2017-09-22
Nest herman/json map based on data-type

### LINKED ISSUE #164: Nest maps by type in herman-add and accoutrement-add JSON.
Right now `herman-add` and `accoutrement-add` output JSON with all the various types (e.g. colors, fonts, sizes) together at the top level.

```json
{
  "component-colors": {
    "text": "#636d73"
  },
  "sans": {
    "name": "Open Sans",
    "stack": ...
  }
}
```

This has the potential for name-conflicts, and also doesn't seem ideal when the JSON is passed other places (e.g. to the App JS) and needs to be accessed. It seems clearer to nest them by type, so the JS could e.g. just grab colors without worrying about fonts.

```json
{
  "colors": {
    "component-colors": {
      "text": "#636d73"
    },
  "fonts": {
    "sans": {
      "name": "Open Sans",
      "stack": ...
    }
  }
}
```

Note: This would necessitate requiring a `type` argument for `herman-add`, as well as changing how the JSON is accessed by Herman: `{{ color_palette(data['colors'][key]) }}`

mirisuzanne:
Fixes #164


scss/config/_requirements.scss
```diff
@@ -0,0 +1,7 @@
+// External Requirements
+// =====================
+
+@import '../../node_modules/accoutrement-color/sass/color';
+@import '../../node_modules/accoutrement-scale/sass/scale';
+@import '../../node_modules/accoutrement-type/sass/type';
+@import '../../node_modules/accoutrement-layout/sass/layout';
```
jgerigmeyer: @mirisuzanne Why are these required now?

jgerigmeyer: I don't see this file being used anywhere, and these same modules are imported elsewhere.

mirisuzanne: oh, I know how that happened… the idea was to move them into one place. I'll fix it.

PR #169, opened 2017-10-17
[fix-fonts] Move font specimens into iframes.

### LINKED ISSUE #76: How shall we handle styleguide/project CSS conflicts?
PR #88, #169

jgerigmeyer:
Fixes #76 and #172.


CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
   [#106](https://github.com/oddbird/sassdoc-theme-herman/issues/106)
 - BREAKING: Group navigation is ordered according to config `groups` order –
   [#150](https://github.com/oddbird/sassdoc-theme-herman/issues/150)
+- BUGFIX: Do not error if `sass` config option is `undefined`.
+- BUGFIX: Do not break layout on fonts with single-item stack –
+  [#172](https://github.com/oddbird/sassdoc-theme-herman/issues/172)
+- BREAKING: `@preview font-specimen` is replaced by new `@font` annotation.
+- BREAKING: Removed `customHead` option; replaced by multiline `@font`
+  annotation.
```
jgerigmeyer: A few subtle things that should be documented related to the `@font` annotation:

1) The font family name will default to the name of the font-map unless a `name` key is found.
2) If the font-map contains an `svgid` key, that value will be used for the `@font-face` `svg-file.svg#svgid` suffix. If none is found, it will default to the font family name.
3) If webfonts require additional HTML (e.g. external Google stylesheet or Typekit JS), they must be initialized with the `@font` annotation for them to work properly in `@example` annotations. The custom HTML defined in the `@font` annotation is automatically inserted into the `@example` annotation `<head>`.

jgerigmeyer:
@mirisuzanne @davisagli @stacyk You can test this on Herman itself as-is, but do the following if you want to test it with a sample project that includes Google, Typekit, and local fonts:

- clone https://github.com/oddbird/herman-test-project
- within that repo, edit the sassdoc theme in `gulpfile.js` (https://github.com/oddbird/herman-test-project/blob/master/gulpfile.js#L45), pointing it to your local copy of Herman with the latest from this branch pulled (e.g. `theme: './../sassdoc-theme-herman/',`)
- in herman-test-project, run `yarn`, then `gulp serve`, and view the project at `localhost:3000`.

templates/item/example.macros.j2
```diff
@@ -49,7 +49,7 @@
       {{ (initial + compiled)|safe }}
     </div>
 
-    {% if example.rendered and not (example.type in css) %}
+    {% if example.iframed and not (example.type in css) %}
       <div class="example-render">
         <iframe class="render-example" srcdoc="{{ example.iframed }}" width="100%" height="0"></iframe>
```
mirisuzanne: Not new, but why the default `height="0"`?

jgerigmeyer: JS sets the iframe height to the iframe `body.scrollHeight` on page load. I think the purpose of the initial `height="0"` is to prevent a flash before the JS sets the new height.

jgerigmeyer:
@mirisuzanne Regarding the console errors, can you see where those are happening exactly? I'm pretty sure they're only on the "examples" page of herman-test-project, which is #174.

index.js
```diff
@@ -600,6 +648,133 @@ herman.annotations = [
...
+   * NOTE: Locally-hosted fonts require global `fontpath` Herman setting.
+   */
+  function font(env) {
+    const valid_formats = ['ttf', 'otf', 'woff', 'woff2', 'svg', 'svgz', 'eot'];
+    // Matches first occurence of text within `''` or `""`
+    const keyRE = /(["'])(?:(?=(\\?))\2.)*?\1/;
+    // Matches text within `()`
+    const variantsRE = /\(([^)]*)\)/;
+    // Matches text within `{}`
+    const formatsRE = /{([^}]*)}/;
```
jgerigmeyer: @davisagli @wlonk Here are the other three REs I've added; happy for feedback.

wlonk: So, there is one thing in these that gives me pause:

`url("http://some/wacky/url/with/a/)/in/it")`

You wanna deal with quoted values differently, you may need a context-free grammar, so I guess it's really "do you even want to deal with that?"

jgerigmeyer: Gah, I didn't think about it that URLs could potentially contain parentheses. So, to deal with this correctly, if the URL is not quoted then I just wait for the first `)`, but if it is quoted I could just wait for the next `")` or `')`, yes?

davisagli: Technically no, since quotes can be escaped (https://www.w3.org/TR/CSS2/syndata.html#value-def-uri). Surely there's an existing library for parsing CSS that we could use instead of discovering the edge cases ourselves?

jgerigmeyer: @davisagli I can look. The regex I'm using (https://github.com/oddbird/sassdoc-theme-herman/pull/177/files#diff-9f97a9c36fbbfb0d19068d6fae3bb84eR85) takes escaped quotes into account.

PR #170, opened 2017-10-17
[ordered-groups] Use group order as set in config.

### LINKED ISSUE #150: I want to define the order of groups/documents in my navigation
Currently I'm using workarounds like this on from Susy…

```yaml
groups:
  a-config: 'Configuration'
  b-api: 'Grid Functions'
  plugin-utils: 'Plugin Utilities'
  plugin_svg-grid: 'Plugin: SVG Grid Image'
  su-math: 'Su: Core Math Engine'
  x-normal: '[Susy Normalization]'
  x-parser: '[Susy Syntax Parser]'
  x-validation: '[Su Input Validation]'
```

jgerigmeyer:
Fixes #150.


index.js
```diff
@@ -217,6 +217,7 @@ const prepareContext = ctx => {
    * templates to manipulate the indexed object.
    */
   ctx.byGroup = byGroup(ctx.data);
+  ctx.groupOrder = Object.keys(ctx.groups);
```
davisagli: This is okay, but I think I would have been more inclined to change `byGroup` to return the groups in the order from `ctx.groups` rather than the order they appear in the data.

PR #177, opened 2017-10-25
Fix url(...) usage in client-provided @example annotation CSS.

### LINKED ISSUE #174: Fix local fonts and background images used in @example annotations.
PR: #177.

Currently, the `customCSS` option is passed into the `@example` annotation `iframe`, but any internal--that is, not http(s)--`url()` paths referenced in that CSS file are broken. This means that local fonts and background images will not work.

Proposal:

Parse through the `customCSS` file, grab any internal path within `url(...)`, copy the file into Herman `assets/`, and replace the path with the correct destination.

TODO:

- [x] make sure `@example` iframe can access local fonts
- [x] make sure `@example` iframe can access background-images referenced in `customCSS`
- [x] do not re-copy fonts if already copied by `@font` annotation

jgerigmeyer:
Fixes #174.


jgerigmeyer:
@davisagli @mirisuzanne This is ready for review. Same functional-review instructions as https://github.com/oddbird/sassdoc-theme-herman/pull/169#issuecomment-338716216.

lib/parse.js
```diff
@@ -73,4 +75,58 @@ module.exports = {
...
     return resp;
   },
+  customCSS: (file, enc, env) => {
+    // Parse the contents of a file containing CSS output. For any occurences of
+    // `url(...)` with relative paths, store the paths (for later copying) and
+    // replace them with the eventual destination paths.
+    env.customCSSFiles = env.customCSSFiles || [];
+    const src_dir = file.path.substring(0, file.path.lastIndexOf('/'));
+    let css = file.contents.toString(enc);
+    const regex = /url\(([^)]*)\)/g;
```
jgerigmeyer: @davisagli @wlonk I generally hate regular expressions, so I'm going to point out a few I've added to Herman recently (this is one), in case either of you sees red flags or has suggestions.

wlonk: That looks short, sweet, and legible to me.

lib/parse.js
```diff
@@ -73,4 +75,58 @@ module.exports = {
...
+      // Ignore quotes, if content is quoted
+      if (src_path.startsWith("'") || src_path.startsWith('"')) {
+        src_path = src_path.substring(1, src_path.length - 1);
+        start_index = start_index + 1;
+      }
+      if (
+        !src_path.startsWith('http') &&
+        !src_path.startsWith('//') &&
+        !src_path.startsWith('data:')
+      ) {
```
jgerigmeyer: @mirisuzanne Do these three cases capture the non-local-path options?

lib/parse.js
jgerigmeyer: Ha! I think "legible" is pretty much out of the question for a regex, but that's my own bias... 😉

lib/parse.js
mirisuzanne: Yeah, I think so. That all makes sense to me.

jgerigmeyer:
@davisagli Our other options are mostly:

1. Write our own postcss plugin (this is what Webpack's css-loader does).
2. Try to use the existing postcss-url plugin (or another similar less-popular postcss plugin).

I played a bit with the second option, but it wasn't straightforward and I'm not really sure what it gains us. For what it's worth, the regex they use breaks on e.g. `url("../(images)/foobar.png")`.

I'd be more willing to spend time on it if we continue running into problems with our custom approach. As it is, I think it's relatively straightforward and seems to be working fine.

davisagli:
Yeah, okay

PR #184, opened 2017-11-08
[deprecate-macro] Remove @macro annotation.

### LINKED ISSUE #182: Deprecate @macro annotation?
Currently, the `@macro` annotation doesn't do anything, because we're not displaying it in the Herman templates. Everywhere we used to use it we've replaced it with `@example njk`. Is there any reason to keep `@macro` around? I guess it's currently the only way to document within a nunjucks file... If we don't deprecate it, we should make sure it displays in the compiled docs.

jgerigmeyer:
Fixes #182.


wlonk:
Why are we removing it? I'm curious.

PR #190, opened 2017-11-14
[icons-without-nunjucks] Handle @icons annotation without requiring Nunjucks.

### LINKED ISSUE #47: Include icons without using nunjucks?
https://oddbird.slack.com/archives/oddbird/p1484675969000035

jgerigmeyer:
Fixes #47.


davisagli:
Overall I like this simplification, but I'm not sure I understand why we still bother to make a minified svg file rather than just including the icons inline in the examples?

Sorry for taking so long to catch up on code review!

jgerigmeyer:
@davisagli That's a good question. I guess I don't know enough about the pros/cons to have a strong opinion. But I think we want to minify them either way, and in this case I'm not sure I see the argument for how inline is actually simpler?

mirisuzanne:
Inline SVG has styling advantages, and is something we may want to move towards on some projects. Otherwise it's a performance question that depends on how many icons, and how often they are used. If you can put an icon in the sprite once, and call it over and over, that can have some advantages. In our previews, that seems like it would be the same either way - so not urgent to change. In our other projects, we should pay closer attention to that decision.

davisagli:
Yeah, I wasn't trying to make a point about what we do "for real" in projects, just about how they get rendered for Herman previews. There's certainly an argument for keeping it as similar to reality as possible. If the approach here works fine than I don't think it's worth revisiting.

PR #198, opened 2017-11-17
[site-search] Wire up search page

### LINKED ISSUE #28: Add Search
Perhaps we should add a search for easier ways to find things.

(This is for the entire styleguide, while #52 helps filtering a single page)

wlonk:
Fixes #28.

seal


jgerigmeyer:
@wlonk This seems like such a common thing; I'm skeptical about the value of writing our own solution. Have you looked into existing (open-source) products? What about something like http://www.tipue.com/?

wlonk:
I did see that, but it seemed that grokking it would take longer than writing our own. Unsure.

In any case, we need to have a page for it, so.

jgerigmeyer:
Documenting some of the options for search plugins:

- http://www.tipue.com/search/docs/
- https://lunrjs.com/
- http://fusejs.io/
- http://elasticlunr.com/

jgerigmeyer:
@mirisuzanne @stacyk This is functional, ready for styling. The packaged CSS (currently not being used) can be found at: https://github.com/Tipue/Tipue-Search/blob/master/tipuesearch/css/tipuesearch.css

mirisuzanne:
@wlonk @jgerigmeyer any way we can improve these result summaries?

summary

> "menu icon menu icon menu menu Herman 1.0.0-alpha.15 menu menu menu menu menu menu Herman 1.0.0-alpha.15 Herman1.0.0-alpha.15 Herman Configuration Changelog Contributing Exporting Styles to ..."

wlonk:
Eeesh. Is that just pulling the top of the page content?

So, per docs:

> `showContext`
> The search term is shown in context in the descriptive text if set to true. The default is true.

So I am surprised at that result.

wlonk:
Ah ha!

> `liveContent`
> In Live mode, the element containing the content to be searched. The default is '*'.
> `liveDescription`
> In Live mode, the element containing the descriptive text. The default is '*'.

Changing one or both of those should help.

jgerigmeyer:
Replaced by #203.

PR #201, opened 2017-11-21
[simple-nav] Single nav, always visible on large screens

### LINKED ISSUE #200: Can we remove the duplicate menus?
Problems:
- We want it visible by default on large screens, hidden on small screens
- We want it to overlay on small screens, and live inline on large screens

Questions:
- Do we need the option to hide at all on large screens? This may be easiest if large screens always show the menu.
- Can we pass the breakpoint to JS via CSS variables? Add `--nav-breakpoint` to the nav menu, and let the JS access it from there?

mirisuzanne:
Fixes #200.

- [x] Miriam: single menu, always visible on large screens
- [x] @jgerigmeyer always change menu-expanded to "true" above `--nav-breakpoint` (set on nav menu), and "false" below.


jgerigmeyer:
@mirisuzanne A number of thoughts/questions:

1) I assume you mean `aria-expanded`, not `menu-expanded`?
2) Do we care about browser support? My understanding is that CSS variables aren't supported in IE. https://caniuse.com/#feat=css-variables
3) Can't we use Herman to export our CSS JSON, and access that from JS to get a CSS-defined breakpoint?
4) Above the breakpoint, it makes sense to force `aria-expanded` to `"true"`. Do we also care about the value of the button `aria-pressed` value?
5) Below the breakpoint, I don't think we want to force `aria-expanded` to `"false"`. If the user has opened the menu (such that the button has `aria-pressed="true"`), then we probably want to leave `aria-expanded="true"` on the nav, yes?

mirisuzanne:
@jgerigmeyer yes, that all sounds right. I wasn't sure if we needed browser-support for variables in order to access that from the CSS. If you prefer JSON, that's fine with me.

jgerigmeyer:
@mirisuzanne Okay, I think this is ready for review. Right now the JS is _not_ updating the button `[aria-pressed]` when above the breakpoint. If it's important for accessibility for that to be updated as well, we can do that. (It was just simpler not to, since I can use that as an indicator for the "canonical" value of the nav `[aria-expanded]` when back below the breakpoint.)

jgerigmeyer:
@mirisuzanne You're the owner of the PR, but I'd appreciate your review on this today when you get a chance. I plan to do another RC release today, after this is merged.

mirisuzanne:
@jgerigmeyer approved!

mirisuzanne:
@jgerigmeyer I don't think the off-sync aria-pressed is an issue.

PR #202, opened 2017-11-21
[brand-cleanup] Cleanup colors and spacing

mirisuzanne:


docs/CHANGELOG.html
```diff
@@ -44,12 +44,12 @@
 
 
 
-      <h1 data-sassdoc="project-title">
+      <div data-sassdoc="project-title">
         <a href="index.html" rel="home" data-sassdoc="project-name">Herman</a>
```
SondraE: Looks odd to me that the link underlines are pale blue (though I do like that the sidebar nav underlines are pale blue). I very much like that the underlines are not bold and obtrusive. What would it look like if the underlines were a very faded version of the red link color? I like the idea of hiding underlines for headings.
Also it's a little odd that links like @font and @preview are blue with a blue underline and then just the underline turns pink. Another thing that happened is I found myself thinking that the blue without underlines were links too. Basically, this whole link situation feels a bit chaotic in terms of color.

docs/CHANGELOG.html
```diff
@@ -58,7 +58,7 @@ <h1 data-sassdoc="project-title">
 
   
     
-  <nav role="navigation" data-nav-menu="small" data-toggle="target" id="nav-small" data-target-id="nav-small" aria-expanded="false" data-auto-closing="false" data-auto-closing-on-any-click="false">
+  <nav data-nav-menu="small" data-toggle="target" id="nav-small" data-target-id="nav-small" aria-expanded="false" data-auto-closing="false" data-auto-closing-on-any-click="false">
 
     <div data-sassdoc="nav-subsection">
```
SondraE: This is not a big deal, but it's not easy to tell the two subtitle sizes apart from one another. Maybe make the smaller one not bold?

docs/CHANGELOG.html
mirisuzanne: Here's a lighter red underline:

link screenshot

The blue isn't about links at all, that's the color we've been using for bits of inline code. When code is also linked, it seems like the code color is taking precedence - which we could fix. If the code style also looks like a link style to you, we should come up with a different way to make code stand out.

mirisuzanne: Here's an attempt at using bold instead of blue for inline code. Too heavy handed?

bold code screenshot

mirisuzanne:
@SondraE I pushed a few changes. Let me know if that addresses your review!

docs/CHANGELOG.html
mirisuzanne: (I switched to a semibold for code. pull the latest to see it)

scss/config/_fonts.scss
```diff
@@ -36,10 +38,8 @@ $sans-font: (
 /// A readable monospace font for code previews.
 ///
 /// @font 'code' (regular, bold)
-///   <link href="https://fonts.googleapis.com/css?family=Source+Code+Pro:400,700" rel="stylesheet">
+///   <link href="https://fonts.googleapis.com/css?family=Source+Code+Pro:400,600" rel="stylesheet">
```
stacyk: Since we changed this to 600, but are still calling "bold" the browser associates that to 700. So are we loading a faux bold here in some browsers? I wonder if we should use 600 instead when we call it later?

mirisuzanne: ok, that makes sense. @stacyk do you want to make that change, so we can get this merged?

jgerigmeyer: If I'm not mistaken, addressing this comment is the only thing remaining on this PR before merge?

mirisuzanne:
@jgerigmeyer @stacyk that should address the font-weight comment. According to [CSS Tricks](https://css-tricks.com/almanac/properties/f/font-weight/) "bolder" will find the next available weight up the stack or fallback to 700, which means it will continue to work no matter what weight we load for bold code.

PR #203, opened 2017-11-22
🐈 [new-site-search] Add custom site search using Lunr.js.

jgerigmeyer:
Another attempt at fixing #28.

zebra


jgerigmeyer:
@mirisuzanne Take a look at the results here... ready for styling if the functionality looks okay. We have more control over the functionality and markup here, so let me know if you want changes.

gulpfile.js
```diff
@@ -421,3 +423,10 @@ gulp.task('imagemin', () => {
 });
 
 gulp.task('minify', ['jsmin', 'svgmin', 'imagemin']);
+
+gulp.task('precompile-templates', () =>
```
mirisuzanne: This had some issues on occasion – I had to restart my server quite often while working.

jgerigmeyer: Hm, what sort of errors?

jgerigmeyer: Only time I've been able to reproduce errors is when I make multiple quick changes in a row, before compilation has time to complete.

jgerigmeyer:
Right now I'm showing 15 words before and after every search result match, but showing all result matches. When you search for a common word (e.g. `color`), that means a lot of matches on some pages. I wonder about limiting it to only display the context for the first 3 or 4 or 5 matches or something like that?

jgerigmeyer:
Okay, I think this is to a point where I'm okay reviewing/merging it. There's still one small bug in the JS search tool that I'm waiting on a fix for (see https://github.com/olivernn/lunr.js/issues/320), but I expect that to come in shortly.

This branch also inspired me to add two new issues (#204, which can probably wait for after 1.0, and #205, which should probably happen ASAP).

`<aside>`
To see the Lunr bug, search for two words, e.g. `color palette` -- the second word is not properly associated with the "field" it was matched in, e.g. "title" or "contents". Right now I'm defaulting to "contents", which means if it was actually matched in the "title" then it gets highlighted in the wrong place. The highlight of `nting co` in the following screenshot should be highlighting `Palettes` in the title.

screenshot 2017-11-23 12 22 51
`</aside>`

jgerigmeyer:
Sidenote: I added another branch, `new-site-search-fuse` (https://github.com/oddbird/sassdoc-theme-herman/tree/new-site-search-fuse), which uses the Fuse.js package instead of Lunr.js. Check it out if you're interested in seeing the difference between two search algorithms. It's much smaller in size and there are a few things I like better about it, but it's much more "fuzzy" and I can't get it to act more like a typical strict search. (There are some relevant issues on the repo that I'm following, e.g. https://github.com/krisk/Fuse/issues/147). So until the results there are better, I think we stick with Lunr.

assets/js/base.js
```diff
@@ -190,6 +190,14 @@ window.Herman = (function base(Herman, $) {
     });
   };
 
+  Herman.getUrlParams = () => {
+    let params = {};
+    if (typeof window.deparam !== 'undefined') {
+      params = window.deparam(window.location.search.substr(1));
+    }
+    return params;
```
wlonk: Do you want to get URL params if `deparam` _is_ `undefined`?

assets/js/base.js
```diff
@@ -209,5 +217,127 @@ window.Herman = (function base(Herman, $) {
...
+                  }
+                  node.nodeValue = text.join(' ');
+                }
+              });
+            },
+          });
+        }
+        // Add search result template to set of results
+        matches = matches.add(el);
+      }
```
wlonk: It's not urgent, but if `showResults` could be decomposed into well-named things, that'd probably help. It's a bit hard to follow.

assets/js/base.js
```diff
@@ -209,5 +217,127 @@ window.Herman = (function base(Herman, $) {
...
+  const doSearch = function(data, val) {
+    // Grab doc store from data
+    Herman.searchStore = data.store;
+    // Initialize Lunr index from precompiled data
+    const idx = window.lunr.Index.load(data.idx);
+    const results = idx.search(val);
+    showResults(results, val);
+  };
+
+  Herman.getSearchData = function getSearchData() {
```
wlonk: Why isn't this an anonymous function assigned to `Herman.getSearchData`?

assets/js/base.js
```diff
@@ -209,5 +217,127 @@ window.Herman = (function base(Herman, $) {
...
+    const params = Herman.getUrlParams();
+    const hasLunr = typeof window.lunr !== 'undefined';
+    const hasNunjucks = typeof window.nunjucks !== 'undefined';
+    // Only fetch search data if on search results page with query term
+    if (params && params.q && hasLunr && hasNunjucks) {
+      const request = new XMLHttpRequest();
+      request.open('GET', '/search-data.json', true);
```
wlonk: Can we not use `fetch` here? Is this a backwards-compatibility thing?

lib/utils/render.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const $ = require('cheerio');
```
wlonk: TIL about cheerio!

assets/js/base.js
jgerigmeyer: We're currently using the `slim` build of jQuery, which does not include `$.ajax` &c. It didn't seem worth adding for simple `$.getJSON` call, so I did it the old-school way.

assets/js/base.js
jgerigmeyer: Good call; no reason.

assets/js/base.js
jgerigmeyer: ¯\_(ツ)_/¯ It's kind of a theoretical question... this fn should never be called currently in cases where $.deparam isn't defined. (There are various pieces of JS that are only used by the search-results page, so only loaded on that page.)

PR #208, opened 2017-11-27
[remove-preview] Replace @preview annotation with @colors/@ratios/@sizes.

### LINKED ISSUE #125: I want a better design for sizing previews
- [x] Ratios (or do we want to remove this preview?)
- [x] Text Sizes (looking the best of what we have)
- [x] Spacing sizes
- [x] Larger sizes (output only)

### LINKED ISSUE #206: `@preview` is no longer universal. Should it be removed?
When we first built it, all previews used the `@preview` syntax. Now only colors & sizes/ratios continue to use that prefix, while `@font` and `@icons` get their own annotation. Is there a reason to hold onto the shared syntax at all, rather than using e.g. `@colors`, `@sizes`, and `@ratios`?

wlonk:
Fixes #206.
Fixes #125.

yak


lib/annotations/sizes.js
```diff
@@ -0,0 +1,31 @@
...
+    let key, value;
+    raw.split(';').forEach(option => {
+      const parts = option.split(':');
+      key = parts[0].trim();
+      value = parts[1] ? parts[1].trim() : null;
+      if (key) {
+        options[key] = value;
+      }
+    });
+    return options;
```
jgerigmeyer: Can we abstract these three cases into a utility fn for converting `key: value; key: value` into an object?

scss/config/_scale.scss
```diff
@@ -56,7 +56,7 @@ $text-sizes: (
 /// A set of standard spacing sizes.
 ///
 /// @group config-scale
-/// @preview sizes; style: ruler
+/// @sizes style: ruler
```
mirisuzanne: Is there a reason we decided to keep using keyword args for style & key, rather than following the syntax used elsewhere?

wlonk: No, the only reason is that I made a minimal change, as at that point I did not yet know how you wanted it to be.

jgerigmeyer: @mirisuzanne What syntax used elsewhere are you referring to? How would you like this to work?

mirisuzanne: I left comments on the slack thread yesterday. Icons and fonts both take a bare key string as the first argument, and fonts take an additional parenthetical arg. That matches more closely with other SassDoc annotations.

For colors/ratios (which only take a key), that might look like:

```scss
/// @colors brand-colors
/// @ratios text-ratios
```

For sizes (which take both key and style), maybe: 

```scss
/// @sizes <key> (<style>)
/// @sizes spacing-sizes (ruler)
```

jgerigmeyer: Hrm, that's the problem with Slack threads -- no one sees them except the author.

Some thoughts:

1) Right now, `@font` expects the key to be within `'` or `"`, but `@icons` does not. Is there any reason for the difference? We should probably standardize, maybe just doing away with quotes entirely.
2) `@ratios` also accepts a `count` arg, not just a `key`.
3) I generally like the syntax you're proposing, but the downside is that all of the arguments are implicit -- you have to know what arguments are expected, in what order. In some ways that seems like a step backward from `key:value` args. For instance, if we expect `@sizes (ruler)` and `@ratios (4)`, it's not clear (without looking at the API documentation) what those arguments mean. But you're correct that it matches SassDoc's API more closely, although I don't see any SassDoc annotations using `(...)` for arguments.

mirisuzanne: 1. SassDoc doesn't use `'` or `"`, so I vote to drop it.
2. 👍 good catch.
3. 
  a. Yeah, I think it would be great if both were available. Is that an option?
  b. It looks like SassDoc uses `{}` for types, `[]` for defaults, and `<>` for links. They sometimes use nothing if they expect e.g. a link first and then optional text. The `style` arg might make sense as a type: `@sizes {ruler} spacing-sizes` (ideally the order doesn't matter). Ratio-count makes less sense there, but I don't like the way our ratio-previews work anyway, so maybe I should just do that redesign, and try to eliminate that argument.

jgerigmeyer: - `1`: Ok, I'll ditch the quotes for `@font`.
- `3a`: Hm, I'm not sure the benefit of supporting two different APIs is worth the added complexity?

SassDoc generally does not allow "any order" for the annotation arguments. I'd propose we follow their lead for these, such that the first word (anything except whitespace) is interpreted as the `key`, and then the args follow. I'll leave `@font` as-is with `()` for variants and `{}` for formats. For `@sizes`, I'll go with `{}` for the style. @mirisuzanne That sound okay?

jgerigmeyer:
I think this is (once again) ready for review. @mirisuzanne, does this now also fix #125?

.gitignore
```diff
@@ -10,4 +10,8 @@ test/js/dest/
 *~
 .livereload
 *.orig
+*.css.map
+*.js.map
 dist/
+docs/assets/
+docs/search-data.json
```
jgerigmeyer: I decided to leave the HTML files in `docs/` checked into the repo, because I often use the diffs of the compiled HTML as a form of functional testing.

.npmignore
```diff
@@ -0,0 +1,12 @@
...
+.cache/
+jscov/
+.nyc_output/
+test/
+*~
+.livereload
+*.orig
+*.css.map
+*.js.map
+.circleci/
```
jgerigmeyer: Normally, an npm package uses `.gitignore` to decide what to include in the packaged version. Since we're now git-ignoring `dist/`, which is required by Herman to work, I added `.npmignore` so that we can define them independently.

templates/item/preview.macros.njk
```diff
@@ -235,66 +229,102 @@
...
+        {% set width = width + 'em' %}
+
+        {{ size_ruler_row(
+          name='—' if ( result == 1 ) else step,
+          value='base' if ( result == 1 ) else result|round(2),
+          width=width
+          ) }}
+      {% endfor %}
+    {% else %}
+      {% for value, items in data|list|groupby('value') %}
```
jgerigmeyer: @mirisuzanne I tried sorting here (see https://github.com/oddbird/sassdoc-theme-herman/commit/159659262982ce9dca73bbb4120e75addc449179), but I think it's unhelpful unless we intelligently sort by calculated size. The simple `|sort(attribute='value')` is too dumb, and just sorts alphabetically. I'd say better to allow the user to sort manually in their Sass map.

jgerigmeyer:
@wlonk You're the creator of the PR originally, but I'd appreciate your review here as well.

templates/item/preview.macros.njk
mirisuzanne: @jgerigmeyer I agree. That's what I saw last night and just assumed I was doing something wrong. I'll rip it out.

mirisuzanne: Oh, you already did. Great.

jgerigmeyer:
@mirisuzanne I see you pushed some cleanup... does that mean you "approve" the PR?

Also, does this close out #125, or is more desired there?

mirisuzanne:
Yes and Yes. I lost battery power and had nowhere to plug in…

PR #209, opened 2017-11-27
[client-side-tests] Add tests for client-side JS.

### LINKED ISSUE #204: Add jstests for client js (assets/js/).
Probably using Karma or Intern or another test runner that can use Chrome Headless?

wlonk:
Fixes #204.

yak


karma.conf.js
```diff
@@ -0,0 +1,59 @@
...
+
+module.exports = function(config) {
+  config.set({
+
+    // base path that will be used to resolve all patterns (eg. files, exclude)
+    basePath: '',
+
+    // frameworks to use
+    // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
+    frameworks: ['jasmine'],
```
jgerigmeyer: Might make sense to use mocha, since we're using it for node tests already?

jgerigmeyer:
@wlonk This is looking good. I did some cleanup to the Karma setup (and getting it reporting correctly on CI). I think now we just need more coverage? You're welcome to work on that, or I can tomorrow.

wlonk:
Oh! Well that might be why my version of it wasn't working.

wlonk:
Coverage is pretty high, but there's one broken test. @jgerigmeyer can you take a look at the `getSearchData` test? Thanks.

.circleci/config.yml
```diff
@@ -22,60 +20,174 @@ jobs:
...
-              ~/.yarn/bin/greenkeeper-lockfile-upload
-            fi
+          command: gulp jstest
+      # Do not report multiple times to coveralls: https://github.com/nickmerwin/node-coveralls/issues/176
+      # - run:
+      #     name: Report Test Coverage [node 6]
+      #     command: nyc report --reporter=text-lcov | coveralls
+      - run:
+          name: Check Test Coverage [node 6]
+          command: nyc check-coverage --statements=100 --functions=100 --branches=100 --lines=100
```
wlonk: As a philsophical point, I feel that anything we can push from `.circleci/config` into in-repo configs for the tools in question, we should. That way, when we run the tools locally, we get the same options set. Of course, Gulp handles a lot of that for us at the moment, which is why I say this is philosophical, but I think worth saying.

test/clientjs/test_base.js
```diff
@@ -220,15 +220,73 @@ describe('initializeToggles', function() {
...
+    expect($.fn.outerHeight).to.have.been.calledTwice;
+
     $(window).trigger('resize');
-    // @@@ This is not a very useful or informative assertion:
-    expect(iframe.get(0).height).to.equal('20');
+
+    expect($.fn.outerHeight).to.have.been.calledThrice;
+
+    iframe.remove();
+    $.fn.outerHeight.restore();
```
wlonk: One thing I noticed (and you'll see I had to write tests this way) is that if the assertion fails, and stub-restoration you do at the end won't be run, potentially causing cascading failures. I'd suggest putting this either in an `afterEach` (probably overkill) or wrapping the test body in a `try` and putting this in a `finally`?

.circleci/config.yml
jgerigmeyer: Yeah, that's a good point. At this point we're only enforcing coverage on CI, and I don't really want to do that locally in any case. But it wouldn't hurt to pull the settings into `gulpfile.js` or another config, so that CI doesn't contain the canonical settings.

test/clientjs/test_base.js
jgerigmeyer: Ah, that's a great suggestion. I saw you'd done the `try/finally` bit, but I didn't register the reason for it. 👍

PR #210, opened 2017-11-28
🐈 [webpack] Add Webpack, Babel, etc.

### LINKED ISSUE #205: Consider browser support for client JS.
Right now, client JS (in `assets/js/`) is using ES6 features (e.g. `const`, `let`, `=>` fns) without a precompilation build step. On quick glance, I'm guessing this will cause errors in IE < 11, for instance. We should consider what browser support we want, and either 1) re-write the JS to be more backwards compatible, or more likely 2) add a build step using Webpack and/or BabelJS or whatnot.

jgerigmeyer:
Fixes #205.

cat


assets/js/init.js
```diff
@@ -1,12 +1,15 @@
...
-    Herman.initializeToggles();
-    Herman.initializeTabs();
-    Herman.initializeIframes();
-    Herman.getSearchData();
-  });
-
-  return Herman;
-})(window.Herman || {}, window.jQuery);
+// Add ES2015 polyfills
+import '@babel/polyfill';
```
wlonk: Are there import side effects here or something?

jgerigmeyer: Yes. Any time you see an `import` without exposing a variable name (e.g.`import 'bar'` vs `import foo from 'bar'`), you can assume there's a side effect or something exposed on the global window. In this case, Babel reads our `.browserslistrc` file to see what browsers we support, and adds the necessary polyfills. See https://github.com/babel/babel/tree/master/packages/babel-preset-env#usebuiltins-entry

PR #220, opened 2017-12-06
Allow nested `///` in SassDoc comments.

### LINKED ISSUE #212: Allow nested `///` in SassDoc comments.
When we want to document how to document something, ideally we could do:

```scss
/// /// @colors
```

But SassDoc strips both sets of `///`, leaving just `@colors` instead of `/// @colors`. Right now we're using a workaround -- `/// // @colors`, which results in `// @colors` -- but that's a bit confusing. Can we submit a PR to SassDoc to make this work?

jgerigmeyer:
Fixes #212.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - NEW: Support SassDoc [`googleAnalytics`][googleAnalytics] and
   [`trackingCode`][trackingCode] options --
   [#215](https://github.com/oddbird/sassdoc-theme-herman/issues/215)
+- CHANGE: Use forked version of CDocParser preserving `///` within comments --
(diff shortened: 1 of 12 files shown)


jgerigmeyer:
Note: If we merge this, it allows us to use `///` in Herman comments internally, but it won't allow users of Herman to do the same unless they use our fork of CDocParser, or our PR gets merged in there.

mirisuzanne:
Is there a way to do it already with a escape character breaking up the `///`? I assume maybe not, since it generally happens in per-formated blocks?

jgerigmeyer:
No, there's not a way to manually escape characters. I asked about adding that to SassDoc in the [issue I filed there](https://github.com/SassDoc/sassdoc/issues/519).

Issue #236, opened 2018-01-16
`herman-export` includes incorrectly escaped quotes in Ruby Sass


chrisfromredfin:
I am very possibly doing something wrong, but I'm officially stuck. :)

Here's what I've done:

- set up `jsonfile: 'exported.json'` in my .sassdocrc
- ran 'sass ./sass/style.scss > exported.json' (my exported.json looks like `/*! json-encode: {"\"colors\"": {"\"demo-colors\"": {"\"brand-blue\"": "#0d7fa5", "\"brand-pink\"": "#e2127a"}}} */`
- ran `node_modules/sassdoc/bin/sassdoc sass`
- my style.scss has the $demo-colors map, the $brand-colors variables (all taken from the docs), a herman-add() call at the bottom, and a herman-export at the bottom.

```
// Colors
/// @colors

$brand-blue: hsl(195, 85%, 35%);
$brand-pink: hsl(330, 85%, 48%);
$demo-colors: (
  'brand-blue': $brand-blue,
  'brand-pink': $brand-pink,
);

// Add stuff to the $herman map.
@include herman-add('colors', 'demo-colors', $demo-colors);

@include herman-export;
```

The error I receive is:

```
Unhandled rejection Template render error: (unknown path)
  Template render error: (/redfin/www/hermanblog.rfslocal.site/node_modules/sassdoc-theme-herman/templates/item/_item.njk)
  Template render error: list filter: type not iterable
    at Object.exports.prettifyError (/redfin/www/hermanblog.rfslocal.site/node_modules/nunjucks/src/lib.js:34:15)
    at /redfin/www/hermanblog.rfslocal.site/node_modules/nunjucks/src/environment.js:486:31
    at eval (eval at _compile (/redfin/www/hermanblog.rfslocal.site/node_modules/nunjucks/src/environment.js:565:24), <anonymous>:74:12)
    at eval (eval at _compile (/redfin/www/hermanblog.rfslocal.site/node_modules/nunjucks/src/environment.js:565:24), <anonymous>:119:12)
    at /redfin/www/hermanblog.rfslocal.site/node_modules/nunjucks/src/environment.js:494:25
    at eval (eval at _compile (/redfin/www/hermanblog.rfslocal.site/node_modules/nunjucks/src/environment.js:565:24), <anonymous>:35:12)
    at /redfin/www/hermanblog.rfslocal.site/node_modules/nunjucks/src/environment.js:538:16
[...]

PR #237, opened 2018-01-16
Add links to jsonfile config where relevant

mirisuzanne:


mirisuzanne:
This is part of addressing: #236

PR #269, opened 2018-03-13
Fix bugs in customCSS url rewriting.

### LINKED ISSUE #265: customCSS url() rewriting is broken on Windows
URLs in `customCSS` that are rewritten, end up broken on Windows. The rewriting logic uses the `path` module, which uses `\` as the separator on Windows and you end up with broken URLs.

### LINKED ISSUE #266: customCSS URL rewriting exemption-check is broken
URLs rewritten inside `customCSS` have an exemption list, but it is incomplete and in part broken.

Current exemption logic is:
```js
if (
  !src_path.startsWith('http') &&
  !src_path.startsWith('//') &&
  !src_path.startsWith('data:')
) {
  // ...
}
```

This has two problems:
- It fails to rewrite relative urls such as `url("http_foo/bar.jpg")` which are most definitely relative, because the starting check for HTTP is overly naive.
- It mistakingly rewrites domain-relative paths such as `url("/foo/bar.jpg")`, because it fails to detect them outright.

You can cover all the necessary cases with a simple regex, afaict:
```js
let regNoRewrite = /^(?:https?\:|data\:|\/)/i;
if ( !regNoRewrite.test(src_path)) {
  // ...
}
```

jgerigmeyer:
- Fixes #265 
- Fixes #266

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 
 ## Unreleased
 
+- 🐛 BUGFIX: Fix bugs in `customCSS` URL re-writing --
+  [#265](https://github.com/oddbird/sassdoc-theme-herman/issues/265),
+  [#266](https://github.com/oddbird/sassdoc-theme-herman/issues/266)
 - 🏠 INTERNAL: Upgrade dependencies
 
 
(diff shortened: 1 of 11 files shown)


mirisuzanne:
Not sure I'm set up to test this functionally.

PR #277, opened 2018-03-28
Upgrade to node-sass 4.8.3.

jgerigmeyer:
Fixes warnings, e.g.:

```
DEPRECATION WARNING: Passing a string to call() is deprecated and will be illegal
in Sass 4.0. Use call(get-function("color")) instead.
```


scss/utilities/_maps.scss
```diff
@@ -81,6 +81,11 @@
 ///   /* #{herman-map-compile($brand-colors, 'color')} */
 @function herman-map-compile($map, $function, $args...) {
   $output: ();
+  $type: type-of($function);
+
+  @if ($type == 'string' and function-exists('get-function')) {
```
mirisuzanne: Because the `$function` value comes from users, they should be responsible for passing in the proper format. If we do it for them, that defeats the purpose of first-class functions. If we're calling this function internally, and we want to upgrade, we should get the function before we pass it in.

jgerigmeyer:
@mirisuzanne Okay, updated this to use first-class functions internally.

scss/config/_colors.scss
```diff
@@ -54,9 +54,9 @@ $brand-colors: (
 /// @colors
 /// @group config-colors
 $neutral-colors: (
-  'light-gray': 'brand-blue' ('tint': 80%, 'desaturate': 80%),
-  'gray': 'brand-blue' ('desaturate': 80%),
-  'black': 'brand-blue' ('shade': 30%, 'desaturate': 80%)
+  'light-gray': 'brand-blue' ('tint': 80%, get-function('desaturate'): 80%),
```
mirisuzanne: Why are tint/shade not captured? I wonder if we want some shorthand for this, as it gets quite bulky… :/

jgerigmeyer: Because of the exception in accoutrement-color, they don't have to be: https://github.com/oddbird/accoutrement-color/blob/master/sass/_utils.scss#L79-L81

mirisuzanne: ah, figured it out.

mirisuzanne: @jgerigmeyer suppose we'll want similar in accoutrement-scale?

jgerigmeyer: I don't think there are any changes needed in accoutrement-scale itself, but users will need similar changes if they pass in functions, yes.

PR #278, opened 2018-03-29
Escape backslashes in herman-export string values.

jgerigmeyer:
This fixes bugs with Windows-style paths, e.g.:

```scss
$local-font: (
  'name': 'rockingham',
  'regular': 'rockingham\\rockingham-regular-webfont',
  'formats': 'woff2' 'woff' 'ttf',
);
```

Will now be output as:

```scss
"regular": "rockingham\\rockingham-regular-webfont"
```

Instead of:

```scss
"regular": "rockingham\rockingham-regular-webfont"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ## Unreleased
 
+- 🐛 BUGFIX: Escape backslashes in herman-export string values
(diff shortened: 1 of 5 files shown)


jgerigmeyer:
Yes; they have to be escaped in strings generally.

PR #286, opened 2018-04-04
Move node-sass to peerDependencies.

jgerigmeyer:
This makes two changes:

1. `node-sass` is moved to `peerDependencies`, so it uses the project's version by default (and warns if the project does not have node-sass installed).
2. Does not even require `node-sass` if the user doesn't use `@example scss` annotations.

We could go a step further and remove `node-sass` from `peerDependencies` entirely, which would remove the warning for users who don't plan to use it.

### DIFF
--- a/scss/config/_abstracts.scss
+++ b/scss/config/_abstracts.scss
@@ -4,6 +4,15 @@
 /// # Herman Config: Private Helpers
 /// @group config-utils
 
+// Functions
+// ---------
+$functions: (
+  'darken': get-function('darken'),
+  'desaturate': get-function('desaturate'),
+  'rgba': get-function('rgba'),
+  'convert': get-function('convert-units')
+);
+
 // Config
 // ------
 /// Internal utility for managing Herman and Accoutrement maps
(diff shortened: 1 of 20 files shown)


jgerigmeyer:
Oh, I also updated to the newer versions of accoutrement-* while I was at it; make sure I did that right?

scss/config/_abstracts.scss
```diff
@@ -4,6 +4,15 @@
 /// # Herman Config: Private Helpers
 /// @group config-utils
 
+// Functions
+// ---------
+$functions: (
+  'darken': get-function('darken'),
+  'desaturate': get-function('desaturate'),
+  'rgba': get-function('rgba'),
+  'convert': get-function('convert-units')
```
jgerigmeyer: Oh, I guess this one isn't really required, since the newer version of accoutrement-scale gets the function internally?

mirisuzanne: It's not required, but the shortcut is nice. Could just use `'convert': 'convert-units'` for that, though.

PR #340, opened 2019-03-18
Moved gulp deps to devDependencies

eFrane:
I noticed that some gulp dependencies were left in `dependencies` instead of `devDependencies`. As there seems to be no harm done in moving them over and as that would reduce the install / dependency auditing footprint of herman, this seems like a good idea.


jgerigmeyer:
While those are gulp packages, they're actually being used in production with vinyl-fs.

- https://github.com/oddbird/sassdoc-theme-herman/blob/b00c9a2956ce2beff3a09f9c661e1ab277f0bf75/lib/utils/render.js#L6
- https://github.com/oddbird/sassdoc-theme-herman/blob/e5d51e13ce9debd61bf66316e6a49334bf4dec5f/lib/renderIframe.js#L5
- https://github.com/oddbird/sassdoc-theme-herman/blob/e5d51e13ce9debd61bf66316e6a49334bf4dec5f/lib/renderIframe.js#L8

Issue #342, opened 2019-09-12
Support sass (dart-sass) module for building scss examples


jcamilleri13:
Herman currently relies on node-sass to compile scss to css, and cannot utilise the somewhat larger feature-set of the canonical sass implementation. Support for the canonical sass would probably also be required for #341. Perhaps we could look at the `package.json` file and pick the correct parser to use, as [sass-loader](https://github.com/webpack-contrib/sass-loader) does?

Issue #341, opened 2019-09-10
Support Sass module syntax


mirisuzanne:
See:
- https://github.com/sass/sass/blob/master/accepted/module-system.md
- http://modules.sass-lang.com/
- npm `sass@1.23.0-module.beta.1`

This will be a breaking change, and require a major release.

PR #346, opened 2020-04-24
Add support for Dart Sass

duggiemitchell:
## Cute animal pic
418352e6ff99a49092c35c63d3c97674--ps

Fixes: #342 
Fixes: #341 

## Description:

- [x] Set node-sass as default compiler
- [x] Add support for Dart Sass
- [x] Update Docs
- [x] Update Changelog


duggiemitchell:
[Update Herman to work with Dart-Sass.](https://trello.com/c/USlvmZry/11-update-herman-to-work-with-dart-sass)

mirisuzanne:
What's the best way to test this? Can we switch the internal herman documentation to use dart-sass, or should I point another project at my local repo on this branch? 

The main potential issue I'm seeing is how we handle `sass.includes` moving forward. `@import` still exists in Dart Sass, but much of it's (Sass) functionality is being deprecated (it will still support plain-CSS imports). I think, in order to support the transition (and the bonus use-case of plain-css imports), we'll need a new `sass.use` array – allowing authors to make the switch explicitly.

docs/config-scale.html
```diff
@@ -988,12 +988,12 @@ <h3 class="item-subtitle">
     </td>
     <td class="size-value">
       <code>
-        0.9rem
+        .9rem
```
jgerigmeyer, line 991: @mirisuzanne I'm not 100% sure what's causing this change in leading zeros... I assume an upgrade in node-sass version?

scss/initial/_root.scss
```diff
@@ -50,7 +50,7 @@ mark {
   #{$link} {
     color: color('action');
     text-decoration-color: color('underline');
-    text-decoration-skip: ink;
+    text-decoration-skip: auto;
```
jgerigmeyer, line 53: @mirisuzanne Note this change -- there was a warning that the spec had changed.

docs/config-scale.html
mirisuzanne: yeah, not sure exactly - but likely just a (non-breaking) change between Sass implementations.

docs/style-banner.html
```diff
@@ -583,15 +583,14 @@ <h3 class="item-subtitle">
 
       
       
-      <pre class="hljs-pre"><code class="lang-css">[data-region=&#39;banner&#39;] {
+      <pre class="hljs-pre"><code class="lang-css">[data-region=banner] {
```
jgerigmeyer, line 586: @mirisuzanne Whoa -- these changes are weird.

Screenshot 2020-04-24 18 30 14

mirisuzanne: Yeah, that's valid - and has the same meaning in CSS. Dart Sass is more aggressive about optimizing output like that.

jgerigmeyer: 🤯

PR #357, opened 2021-05-07
Upgrade dependencies and fix cheerio bug.

jgerigmeyer:


lib/utils/render.js
```diff
@@ -24,12 +24,13 @@ module.exports = function render(nunjucksEnv, tpl, dest, ctx, rendered) {
...
           if (rendered) {
-            let title = $('title', html).first().text().trim();
+            const $ = cheerio.load(html);
+            let title = $('title').first().text().trim();
             // Strip " | <name> Documentation" from page titles
             if (title.includes(' | ') && title.endsWith(' Documentation')) {
               title = title.substring(0, title.indexOf(' | '));
             }
-            const contents = $('[data-page]', html)
+            const contents = $('[data-page]')
```
jgerigmeyer, line 33: @mirisuzanne This was the relevant change. We were not pinning to an exact version of cheerio (`^1.0.0-rc.5`), and they introduced a breaking change in `1.0.0-rc.9`. This fixes that change, and pins to the exact version (until they do a full release).

mirisuzanne (approved):

PR #363, opened 2021-06-18
Upgrade deps.

jgerigmeyer:


jerivas (approved):
Just want to make sure I didn't miss anything important:

- markdown plugin now adds `tabindex=-1`
- imports are now sorted
- external sass deps are silenced (I think?)

jgerigmeyer:
>     * markdown plugin now adds `tabindex=-1`

👍

>     * imports are now sorted

👍

>     * external sass deps are silenced (I think?)

Well, I wish that were true. Unfortunately the new `quietDeps` setting doesn't seem to actually work with Yarn v2, and I tried webpack's `ignoreWarnings` option with no success either (I think because the warning is not being emitted by a webpack loader itself). So this kind of punts on that whole question for now -- hopefully we can update [accoutrement](https://github.com/oddbird/accoutrement/issues/73) directly, which would be the real fix.

mirisuzanne (commented):
I'm curious why we would want `tabindex='-1'` on all our headings? Is that documented somewhere?

jgerigmeyer:
> I'm curious why we would want `tabindex='-1'` on all our headings? Is that documented somewhere?

@mirisuzanne That change came from a newer version of [markdown-it-anchor](https://github.com/valeriangalliat/markdown-it-anchor). We can disable that if we want, but it's the new default setting. See [this issue](https://github.com/valeriangalliat/markdown-it-anchor/pull/86) for discussion.

I don't really feel strongly about it. In fact, given that my goal here was really about silencing the dart-sass deprecation warnings and I wasn't able to do that, I don't know that this PR warrants a new release yet (though it doesn't hurt to merge it).

mirisuzanne (approved):
@jgerigmeyer got it, thanks. LGTM

PR #376, opened 2022-08-11
Disable SassDoc annotation autofilling by default.

jgerigmeyer:
## Description

- By default, disables all autofilled SassDoc annotations (can be overridden with `HERMAN_ENABLE_AUTOFILL` env var).


jgerigmeyer:
[Should Herman disable Sassdoc autofilled annotations?](https://trello.com/c/qOC7nIsL/91-should-herman-disable-sassdoc-autofilled-annotations)

jerivas:
Not sure what autofill we're talking about, but the new env var and docs look good for someone with the appropriate context 👍

jerivas (approved):

jgerigmeyer:
> Not sure what autofill we're talking about, but the new env var and docs look good for someone with the appropriate context 👍

@jerivas Brief version: SassDoc attempts to automatically detect when one Thing (e.g. a function) depends on another Thing (e.g. a variable), without the user needing to add a `@requires` annotation or anything. This is nice in theory, but SassDoc also connects a comment to a Thing even if there's a newline in between, whereas Herman does not allow that. So in this case:

```scss
/// My SassDoc comment...

@function my-fn() {
  @return $my-var;
}
```

SassDoc would see the comment as connected to the fn, and would say hey that fn requires the `$my-var` variable! But Herman would say nope, that comment is unattached from any code, and doesn't require anything. Then things break.

This PR tells SassDoc not to auto-add any annotations when using Herman.

jerivas:
Ah, makes sense, thanks for explaining 👍

scss/samples/_mixins-functions.scss
```diff
@@ -7,7 +7,32 @@
...
+/// To allow autofilled annotations, set the `HERMAN_ENABLE_AUTOFILL`
+/// environment variable to any value before running SassDoc.
+///
+/// Prose blocks can also include examples, and other annotations.
+/// @example scss
+///   .private::after {
+///     content: samples.$private;
+///   }
+/// @link http://sassdoc.com/annotations/
+/// @link http://sassdoc.com/configuration/#autofill
```
mirisuzanne, line 35: These examples aren't very clear in the rendered version. I got to them and wondered if I was seeing a bug. I think we would need to show an escaped version of the input, and then the rendered annotations would make more sense. I guess that's not new, though - just moved.

mirisuzanne (commented):
This seems pretty drastic, if there's a way to just turn off annotations _when there is a line break_. But maybe that's significantly harder?

I'm ok with it, but I don't love it.

scss/samples/_mixins-functions.scss
jgerigmeyer: @mirisuzanne Addressed in https://github.com/oddbird/sassdoc-theme-herman/pull/376/commits/b05d6dfca0a5ca42d84f13595eea8a5f25c2f6f2. What do you think?

jgerigmeyer:
> This seems pretty drastic, if there's a way to just turn off annotations _when there is a line break_. But maybe that's significantly harder?
> 
> I'm ok with it, but I don't love it.

@mirisuzanne Hrm, this is a tough one. There really isn't a way for us to even know what annotations have been autofilled vs not after the fact -- and some annotations like `@require` also impact other items (e.g. the required item) -- so we can't easily just strip these out in prose cases. I do think that the use for autofilling breaks down as soon as Herman introduces new rules for the relationship between a comment and code. So I'd say the options are to do this (and provide an opt-out for users who really want it), or to revisit our rules for prose comments. (I could be convinced to just add a prominent warning about autofilling, but the error is so vague that I'm not sure most users would realize what the underlying issue is.)

mirisuzanne:
Yeah, ok, that makes sense. Let's go with it.

jgerigmeyer:
> So I'd say the options are to do this (and provide an opt-out for users who really want it), or to revisit our rules for prose comments.

@mirisuzanne I guess instead of removing `autofill` from the core annotations, we could try first running the item through our prose logic check, and then then skipping the autofill entirely if we think it's a prose item? Not sure if it would work, but I can give it a shot.

jgerigmeyer:
@mirisuzanne Okay, the changes in https://github.com/oddbird/sassdoc-theme-herman/pull/376/commits/d97aaef2a3f1075ee80ec881924f24b51b79165d now mean that we're only skipping the default `autofill` functionality for items that Herman will consider a "prose" block. Does this seem like a better approach?

lib/annotations/access.js
```diff
@@ -0,0 +1,41 @@
...
+
+      if (env.privatePrefix === false) {
+        return undefined;
+      }
+
+      let testFunc = defaultPrivatePrefixTest;
+
+      if (typeof env.privatePrefix !== 'undefined') {
+        testFunc = RegExp.prototype.test.bind(new RegExp(env.privatePrefix));
+      }
```
jgerigmeyer, line 32: At the point where Herman needs to override this `autofill` fn, we don't have access to the actual `env` yet. Because this annotation relies on values from the runtime `env`, we need to pull this annotation into our repo and rewrite it (instead of just augmenting the SassDoc annotation).

mirisuzanne:
> @mirisuzanne Okay, the changes in [d97aaef](https://github.com/oddbird/sassdoc-theme-herman/commit/d97aaef2a3f1075ee80ec881924f24b51b79165d) now mean that we're only skipping the default `autofill` functionality for items that Herman will consider a "prose" block. Does this seem like a better approach?

Great! Yeah, I like that it's a much more targeted fix to the issue.

PR #383, opened 2022-09-02
Render colors in iframe with custom CSS 

dvdherron:
## Steps to test/reproduce

- Navigate to `/demo_colors` and inspect the color previews.
- the iframe should be visible and show that it's loading custom css 
_Please explain how to best reproduce the issue and/or test the changes locally (including the pages/URLs/views/states to review)._

## Show me
_Provide screenshots/animated gifs/videos if necessary._

### DIFF
--- a/lib/annotations/colors.js
+++ b/lib/annotations/colors.js
@@ -1,11 +1,45 @@
 'use strict';
 
+const Promise = require('bluebird');
+
+const renderIframe = require('../renderIframe');
+const ensureSassJson = require('../utils/ensureSassJson');
+
 /**
  * Custom `@colors` annotation. Accepts optional
  * map-variable name (used to access data from Sass JSON).
  */
-module.exports = () => ({
+module.exports = (env) => ({
   name: 'colors',
   multiple: false,
   parse: (raw) => ({ key: raw.trim() }),
+  resolve: (data) => {
+    const promises = [];
+    data.forEach((item) => {
+      if (!item.colors) {
+        return;
+      }
+      let promise = ensureSassJson(env, '@colors');
+      promise = promise.then(() => {
+        const key =
+          item.colors.key ||
+          (item.context && item.context.origName) ||
+          (item.context && item.context.name);
+        item.colors.key = key;
+        const colorsData =
+          env.sassjson && env.sassjson.colors && env.sassjson.colors[key];
+        if (!colorsData) {
+          env.logger.warn(
+            `Sassjson file is missing colors "${key}" data. ` +
+              'Did you forget to `@include herman.add()` for these colors?',
+          );
+          return Promise.reject();
+        }
+        return Promise.resolve();
+      });
+      promise = promise.then(() => renderIframe(env, item, 'colors'));
+      promises.push(promise);
+    });
+    return Promise.all(promises);
+  },
 });
--- a/lib/renderIframe.js
+++ b/lib/renderIframe.js
@@ -20,6 +20,11 @@ module.exports = (env, item, type) => {
   let includeSassJSON = false;
   const promises = [];
   switch (type) {
+    case 'colors':
+      shouldRender = item.colors;
+      includeSassJSON = true;
+      includeCustomCSS = true;
+      break;
     case 'example':
       shouldRender = item.rendered;
       includeCustomHTML = true;
@@ -116,6 +121,9 @@ module.exports = (env, item, type) => {
 
       let tpl;
       switch (type) {
+        case 'colors':
+          tpl = templates.colors_iFrame;
+          break;
         case 'example':
           tpl = templates.example_iFrame;
           break;
--- a/lib/utils/templates.js
+++ b/lib/utils/templates.js
@@ -5,6 +5,7 @@ const path = require('path');
 const nunjucks = require('nunjucks');
 
 const baseDir = path.resolve(__dirname, '../../templates');
+const colors_iFrameTpl = path.join(baseDir, 'colors', 'base.njk');
 const docTpl = path.join(baseDir, 'doc.njk');
 const example_iFrameTpl = path.join(baseDir, 'example', 'base.njk');
 const fontFaceTpl = path.join(baseDir, 'fonts', 'font_face.njk');
@@ -19,6 +20,7 @@ const sizes_iFrameTpl = path.join(baseDir, 'sizes', 'base.njk');
 
 const templates = {
   baseDir,
+  colors_iFrame: colors_iFrameTpl,
   index: indexTpl,
   search: searchTpl,
   group: groupTpl,
--- a/templates/colors/base.njk
+++ b/templates/colors/base.njk
@@ -23,6 +23,7 @@
   {% endif %}
 </head>
 <body data-herman-preview='colors'>
-  {{ prevew.color_palette(data=sassjson.colors[args.key]) }}
+
+
 </body>
 </html>
--- a/templates/colors/base.njk
+++ b/templates/colors/base.njk
@@ -1,0 +1,28 @@
+{% import 'item/preview.macros.njk' as preview %}
+
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <base target="_parent">
+
+  <title>colors</title>
+
+  {% if customCSS and customCSS.url %}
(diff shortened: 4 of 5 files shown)


dvdherron:
[Render colors in Herman in an iframe with custom CSS (like we do for sizes, ratios, icons, examples, fonts)](https://trello.com/c/6W6pXJtv/93-render-colors-in-herman-in-an-iframe-with-custom-css-like-we-do-for-sizes-ratios-icons-examples-fonts)

dvdherron:
@jgerigmeyer I tried to follow how `ratios` and `sizes` are set up to show iframe previews and I think I have the iframe loading for colors.

I _think_ what's left is to tweak the templates to get the iframe showing in the right place. In `/templates/colors/base.njk` should I be calling the `color_palette` macro in the `<body>`? And in `preview.macros` move the call to the `show.item_section` macro from the `color_palette` macro to a new `colors` macro that renders the iframe?

jgerigmeyer:
> I _think_ what's left is to tweak the templates to get the iframe showing in the right place. In `/templates/colors/base.njk` should I be calling the `color_palette` macro in the `<body>`? And in `preview.macros` move the call to the `show.item_section` macro from the `color_palette` macro to a new `colors` macro that renders the iframe?

@dvdherron Yes, I think that's exactly right! This PR is looking great so far. You'll want the new `colors/base.njk` template to call the macro which actually renders the colors (e.g. `color_palette`), and `show_previews` should instead now render `item_section` with an iframe. You can probably continue following size/ratio as an example.

jgerigmeyer:
> Any idea why that `colors` filter is not being found?

@dvdherron That filter was being added _after_ the `@colors` annotation was resolved -- I fixed that in https://github.com/oddbird/sassdoc-theme-herman/pull/383/commits/e0bef6ce70ccf36e5665f437b2193ffe06c8b116 by using a consistent nunjucks environment throughout.

dvdherron:
> > Any idea why that `colors` filter is not being found?
> 
> @dvdherron That filter was being added _after_ the `@colors` annotation was resolved -- I fixed that in [e0bef6c](https://github.com/oddbird/sassdoc-theme-herman/commit/e0bef6ce70ccf36e5665f437b2193ffe06c8b116) by using a consistent nunjucks environment throughout.

Aha, I was thinking it had something to do with things not being called in the right order. Thanks for the fix 🙏🏽

mirisuzanne (approved):
Everything here looks good. Can we add examples of this being used to render custom properties as colors? Or is that a separate feature, that still needs to be added on top of this one?

jgerigmeyer (approved):
🚀

dvdherron:
@mirisuzanne @stacyk 

I've added an example of a color preview using custom properties:

bild

The one thing I wasn't sure about is if I'm calling the `colors--` mixin from the correct place. Is there a better place to generate the color custom properties than in `/previews/color`?

jgerigmeyer (approved):

scss/samples/_colors.scss
```diff
@@ -92,6 +92,28 @@ $demo-colors: (
...
+/// can also be previewed as palettes.
+///
+/// [ac]: https://www.oddbird.net/accoutrement
+///
+/// @colors
+/// @group demo_colors
+$custom-property-colors: (
+  'accent': var(--callout),
+  'link': var(--action),
+  'btn-text': var(--contrast-light),
```
mirisuzanne, line 109: Huh, the result here is a bit confusing, but technically accurate. Maybe needs a bit of thought about what we're actually trying to show here. 

- Technically, there's a token named `accent`, and it has a value of `var(--callout)`. That's all displayed accurately. But…
- When we use a setup like this, we expect to generate a custom property for each token. If you want to use the `accent` token, you call `var(--accent)`. The current output confused me at first, because `--accent` isn't used anywhere.

I don't know if/how we can or should represent that. At this point, Herman doesn't know that `var(--accent)` will be generated, it only sees a token with a value - which we are displaying correctly.

I'd be interested in @stacyk's thoughts on this. At the minimum, we might want some more text that explains what's happening here?

mirisuzanne (approved):

scss/samples/_colors.scss
dvdherron: @stacyk @mirisuzanne 

Would it clear things up if I explain that the keys ('accent', 'link', 'btn-text') are example tokens based on tokens that were turned into custom properties in a separate location? 

My intention was to show previews from a map that uses uses custom properties instead of raw color values. Would it be less confusing if I used custom properties of tokens that we are already using on the page (i.e. `brand-blue`, `brand-pink`)?

stacyk: I think that using the brand colors that were defined earlier on this page provides clarity. I don't think the name you give them is something we have to be using necessarily:

Screen Shot 2022-09-20 at 12 46 06 AM

Although, I do think we darken our brand pink at some point so the example above is showing the darkened pink on bottom and the regular pink above which is confusing. 

**Retreat Review**

In general, I feel like reading through a few pages of the current documentation, I would love to see more examples or perhaps visuals/graphs of how some things work/look, etc. 

Would it be confusing or appropriate anywhere to show how we are defining custom properties as sass variables in their own partial without a map and provide examples of what we do to turn it into a map/json data for Herman? (Maybe that is more of an Accoutrement example?)

Earlier on this page, it says "Whether you store colors as individual variables or group them into a map, you can use Herman to document and display color palettes." Do we show any examples of storing individual colors and displaying the palettes?

stacyk (commented):
Nice work @dvdherron! I'm excited we have swatches again! I don't really have any major change requests here in this specific PR other than perhaps what we discussed in the custom properties map demo. I added a few notes about retreat talking points but those shouldn't hold up this PR.

scss/samples/_colors.scss
dvdherron: For clarity's sake, I used the `brand-colors` custom properties here (with the color difference explanation @stacyk ). I also tried to explain a bit more about all that's going on here. 

image


Is this more clear?

scss/samples/_colors.scss
```diff
@@ -92,6 +92,28 @@ $demo-colors: (
...
 );
 
+/// ## CSS custom properties
+///
+/// At OddBird, we convert color tokens into
+/// CSS custom properties using [Accoutrement][ac].
+/// Color values stored as custom properties
+/// can also be previewed as palettes.
+///
+/// [ac]: https://www.oddbird.net/accoutrement
```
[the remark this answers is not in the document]
dvdherron: added a more more targeted link to the [colors variables](https://www.oddbird.net/accoutrement/docs/color-vars.html#mixin--colors--) page in Accoutrement.

scss/samples/_colors.scss
```diff
@@ -92,6 +92,28 @@ $demo-colors: (
   'brand-pink': $brand-pink,
 );
 
+/// ## CSS custom properties
```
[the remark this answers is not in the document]
jgerigmeyer: > I wonder if we want to include that Herman will display the css custom property name like var(--callout) in this example. Is that by default as well?

Hm, yeah -- I'm not sure there's an easy way for us to convert the css custom property into other color formats at compile time. The only thing I can think of is running a script at runtime to get the computed `background-color` value from the swatch. So by default it will only ever show the property name, not a color value. 🤷

scss/samples/_colors.scss
mirisuzanne: Much more clear, and good thoughts @stacyk. Would be good conversations for retreat.

scss/samples/_colors.scss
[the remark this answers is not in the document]
stacyk: @jgerigmeyer I don't think it is a dealbreaker to not show color value, but just a bit of a helpful tidbit of info since some colors are pretty similar. 

I'm thinking we can re-word the description of what we show as we say name (accent, in the above example) and then the color in hsl/a, hex, rgb/a... which in the variable example we only show color name and variable name.

stacyk (approved):
The changes you made are great @dvdherron!

scss/samples/_colors.scss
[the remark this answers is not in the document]
jgerigmeyer: @stacyk Added a note in https://github.com/oddbird/sassdoc-theme-herman/pull/383/commits/0433eb634c297e8506558941354f74db2ae1201d

jgerigmeyer (approved):

PR #384, opened 2022-09-08
Replace TinyColor with Colorjs.io

jgerigmeyer:
TinyColor does not support colors that use hue angle notation (e.g. `hsl(53deg 89% 96%)`), so when Sass started adding that by default, we lost the display values for those colors.


test/js/testTemplates.js
```diff
@@ -31,24 +33,29 @@ describe('isString filter', () => {
...
   });
 
   it('switches on formats', function () {
     const actual = this.colors('#fefced');
     const expected = {
       hex: '#fefced',
-      rgb: 'rgb(254, 252, 237)',
-      hsl: 'hsl(53, 89%, 96%)',
+      rgb: 'rgb(99.608% 98.824% 92.941%)',
+      hsl: 'hsl(52.941 89.474% 96.275%)',
```
jgerigmeyer, line 58: @mirisuzanne What sort of precision do we want here? By default colorjs.io displays 5 significant digits, but we can change that.

test/js/testTemplates.js
```diff
@@ -59,10 +66,10 @@ describe('makeNunjucksColors', () => {
         displayColors: ['rgba', 'hsla'],
       },
     });
-    const actual = colors('#fefced');
+    const actual = colors('hsl(53deg 89% 96% / 0.5)');
     const expected = {
-      rgb: 'rgb(254, 252, 237)',
-      hsl: 'hsl(53, 89%, 96%)',
+      rgb: 'rgb(99.56% 98.729% 92.44% / 0.5)',
+      hsl: 'hsl(53 89% 96% / 0.5)',
```
jgerigmeyer, line 72: @mirisuzanne When users enter `'hsla'` or `'rgba'` as the display format, would you prefer that we display like this, or instead e.g. `hsla(53, 89%, 96% , 0.5)` and `rgba(99.56%, 98.729%, 92.44% , 0.5)`? The former is using the regular rgb/hsl formats with colorjs, the latter is using their rgba/hsla formats.

jgerigmeyer:
[Bug with color conversion.](https://trello.com/c/2vdPENpN/97-bug-with-color-conversion)

test/js/testTemplates.js
mirisuzanne: I think that's probably fine for now. Their picker allows for setting that as a user, which we could consider as a future story. https://colorjs.io/apps/picker/oklch

jgerigmeyer:
Shucks, it looks like colorjs.io doesn't currently support Node v14, but they plan to:

- https://github.com/LeaVerou/color.js/issues/206
- https://github.com/LeaVerou/color.js/issues/205

We could wait for that, contribute a PR there, or drop Node v14 support...

test/js/testTemplates.js
mirisuzanne: Are we providing separate formats for rgb vs rgba? Their default is the proper modern standard format, so great to prioritize - the latter would only be useful if we're wanting to give people a 'legacy output' option for backwards compatibility.

mirisuzanne:
> Shucks, it looks like colorjs.io doesn't currently support Node v14, but they plan to:
> 
>     * [Add CI to run tests for old Node.js versions LeaVerou/color.js#206](https://github.com/LeaVerou/color.js/issues/206)
> 
>     * [Avoid Nullish coalescing operator to support old Node.js LeaVerou/color.js#205](https://github.com/LeaVerou/color.js/issues/205)
> 
> 
> We could wait for that, contribute a PR there, or drop Node v14 support...

I don't really know the tradeoffs. How difficult to submit that PR?

test/js/testTemplates.js
jgerigmeyer: We allow users to enter `'rgba'` or `'hsla'` as display formats, but we currently treat them identically to `'rgb'` and `'hsl'` -- but we could adjust that to use the alternate formats I mentioned above.

jgerigmeyer:
> I don't really know the tradeoffs. How difficult to submit that PR?

https://github.com/LeaVerou/color.js/pull/215

test/js/testTemplates.js
jgerigmeyer: @mirisuzanne Preference on the above?

mirisuzanne: The advantage is potentially that it lets you get a legacy format - but that wouldn't be entirely clear from simply the input. So maybe I lean towards no change, and we can provide a way to copy-out legacy formats as a distinct feature, maybe as part of support for gamut-mapping.

mirisuzanne (approved):

PR #385, opened 2022-11-02
Automatically publish documentation

jerivas:
## Description
Use GH Actions to automatically build and publish documentation on every release.

1. On each stable release the `publish-docs.yml` workflow is called
2. `yarn build` outputs to the `docs/` folder
3. The `docs/` folder is pushed to the `oddleventy-docs` branch
4. Netlify picks up the new commit on this branch and re-deploys the site at https://herman-docs.netlify.app/herman/docs
5. oddbird.net proxies the Netlify site under the `/herman/docs/` path (after we merge https://github.com/oddbird/oddleventy/pull/268)

I also added the option of running the workflow on demand. This should allow us to rebuild docs without having to create a release.


jerivas:
Now that I see it here, I wonder if we should just have a separate `docs` branch on this repo and have netlify publish that, instead of a separate repo and netlifly site 🤔

jgerigmeyer:
> Now that I see it here, I wonder if we should just have a separate `docs` branch on this repo and have netlify publish that, instead of a separate repo and netlifly site 🤔

Hm, good question. I can see how that would be cleaner, to not have extra repos. 🤔 I'm fine with whatever you prefer here -- this repo does already have one Netlify deployment (https://herman-staging.netlify.app/), but if it's easy to add a different deployment for a `docs` branch then I'm open to that.

jerivas:
I think it depends on what we want to do. If it's useful to have git history for the built docs, then a separate repo makes sense. If not, then we can just have a separate branch that we only push to when new releases happen.

jgerigmeyer:
> I think it depends on what we want to do. If it's useful to have git history for the built docs, then a separate repo makes sense. If not, then we can just have a separate branch that we only push to when new releases happen.

Won't the branch provide a git history too?

jerivas:
I was thinking that  `docs` is ignored in this repo, but the branch can probably have its own `.gitignore` not related to `main`. I'll explore that

jerivas:
I switched to using a new branch instead of a new repo in 9429e995194a2efa5922be9fe431198d024f9290. Everything else is pretty much the same

jgerigmeyer (approved):
Does this work without modifications to the `.gitignore` file, which ignores `docs/`?

jerivas:
It works if the other branch is not related to main, so it can have its own set of ignored files. I already pushed a branch like that for this repo

jgerigmeyer:
Ah, I see -- so it just has no .gitignore file at all on that branch 👍

jgerigmeyer:
Oops, I just did: #387

PR #387, opened 2022-11-03
Fix docs workflow

jgerigmeyer:

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -43,7 +43,8 @@ jobs:
           SHA=$(git rev-parse HEAD)
           cd docs-branch
           rm -rf herman/docs
-          cp -r ${{ github.workspace }}/docs herman/
+          mkdir -p herman/docs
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -43,7 +43,8 @@ jobs:
           SHA=$(git rev-parse HEAD)
           cd docs-branch
           rm -rf herman/docs
-          cp -r ${{ github.workspace }}/docs herman/
+          mkdir -p herman/docs
+          cp -r ${{ github.workspace }}/docs/ herman/
```
jgerigmeyer, line 47: @jerivas I couldn't get `cp -r` to copy the top-level `docs` directory locally 😕

jerivas: That's weird, both versions work for me as expected. That is, the `docs` folder is copied inside the `herman` folder even when `herman` is completely empty. Let's try this one and hopefully that fixes it.

jerivas (approved):

PR #389, opened 2022-11-08
Fix styles for color previews.

jgerigmeyer:


jgerigmeyer:
[Bug: Herman color preview styles are not loaded in color preview iframes.](https://trello.com/c/jvd58ABb/108-bug-herman-color-preview-styles-are-not-loaded-in-color-preview-iframes)

scss/iframes.scss
```diff
@@ -4,3 +4,4 @@
 @use 'utilities';
 @use 'config';
 @use 'iframes/index';
+@use 'initial';
```
dvdherron, line 7: @jgerigmeyer This was what was causing the wrong fonts to load in the previews. 

They are loading correctly now:

image

scss/iframes/_index.scss
```diff
@@ -1,3 +1,4 @@
 @forward 'base';
 @forward 'font';
 @forward 'size';
+@forward 'color';
```
dvdherron, line 4: Since all the code in this file was only being used in iframes, I moved this file here with the rest of the iframe preview files.

dvdherron (approved):

templates/colors/base.njk
```diff
@@ -11,7 +11,7 @@
 
   {% if customCSS and customCSS.url %}
     <!-- Custom CSS -->
-    <link rel="stylesheet" href="{{ customCSS.url }}" />
+    {# <link rel="stylesheet" href="{{ customCSS.url }}" /> #}
```
jgerigmeyer, line 14: Todo: Revert this before merge.

scss/iframes.scss
mirisuzanne: If this is being loaded in all example output iframes, it's probably a bit much. I think we have two kinds of iframe:

- the kind where we're generating the visualization, with minimum user styles (like these color swatches) - in which case we need the Herman fonts, and these initial styles make sense to include.
- the kind where authors are generating the output, and we should add as little code as possible. There's no reason to include our fonts or any other initial styles in that output.

jgerigmeyer: @mirisuzanne We currently don't load any Herman styles into `@example` iframes, but we have a specific `iframes.scss` file that we load for `@colors`, `@fonts`, `@sizes`, and `@ratios`. (Then we load _all_ Herman styles for `@icons` iframes.)

I was noting that those other annotations (fonts, sizes, ratios) have a limited set of styles (incl Herman fonts etc) already in `iframes/_base.scss`, so I think we maybe want to use that same approach for `@colors` rather than loading in `initial`?

dvdherron: @jgerigmeyer Ah, I see now. The markup for the color previews is missing the iframe-specific selectors that are already in `iframes/_base.scss`. No need to load all of 'initial'. I'll clean that up.

dvdherron: @jgerigmeyer I edited the color template to use the existing styles in `iframes/_base`. Removed some extra whitespace that was happening for colors that had multiple `color-name`'s too.

dvdherron:
> Looks good!
👍🏾
> 
> 
>     1. Do we want to apply a `margin: 0;` to `[data-herman-preview='colors']`? Currently there's a slight margin:
> 
>     2. Extremely minor, but to me it looks like the color name could use a slight top margin or padding -- it seems a bit close to the color preview 🤷

@jgerigmeyer 
Fixed the spacing here to 1) remove the margin around the colors preview and 2) add some extra space above the color names:

bild

templates/item/preview.macros.njk
```diff
@@ -48,19 +48,19 @@
         </div>
         <div class="color-info">
           <strong class="color-name">
```
mirisuzanne, line 50:
```suggestion
          <strong class="color-name" data-herman-label>
```

mirisuzanne (approved):
Looking good - just suggested one (2-part) adjustment

scss/iframes/_base.scss
```diff
@@ -1,5 +1,6 @@
 [data-herman-label],
-[data-herman-th] {
+[data-herman-th],
+.color-name {
```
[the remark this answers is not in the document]
jgerigmeyer: I wondered about that too. Seems like a good idea, tho it will require a few more style adjustments:

Screen Shot 2022-11-10 at 6 10 35 PM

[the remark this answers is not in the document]
dvdherron: Makes sense to just use the `[data-herman-label]` here. fixed in https://github.com/oddbird/sassdoc-theme-herman/pull/389/commits/eabc826c15b8e1da75a529f7985f19204aaaa67f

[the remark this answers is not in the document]
jgerigmeyer: @dvdherron I like this visually, but I think we should probably replace the other `.color-*` selectors with `[data-herman-*]` selectors in `iframes/_color.scss` as well, right? That way the stylesheet we're loading in iframes has herman-specific selectors, and we don't risk potential conflicts with project styles. (This also fits the pattern we're using in the other iframe stylesheets, e.g. `_font.scss` and `_size.scss`.)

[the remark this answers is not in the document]
dvdherron: Agreed. Fixed these in https://github.com/oddbird/sassdoc-theme-herman/pull/389/commits/260cbb785d2743a0da4f082ce6358e9c893b0b76

jgerigmeyer (commented):
LGTM! @mirisuzanne?

mirisuzanne (approved):

PR #390, opened 2022-11-15
[Bug] Herman preview styles overriden by custom styles

dvdherron:
## Description

- Fixes styles for code examples being overriden by site custom styles. 
- Limits the amount of Herman color custom properties generated for color previews. 

## Steps to test/reproduce
- To see a live version of the conflicts, see the color previews on [this OddSite PR](https://deploy-preview-276--oddleventy.netlify.app/styleguide/config-color) using the latest version of Herman.  
- the font sizes are too big and the some of the Herman styles for code examples are being overrwritten by custom styles from OddSite. 
_Please explain how to best reproduce the issue and/or test the changes locally (including the pages/URLs/views/states to review)._

## Show me

Color previews now only contain two generated custom properties: `--herman-brand-blue` and `--herman-brand-pink`:

bild

Previously we were generating custom properties for all Herman colors:

Old:

bild

_Provide screenshots/animated gifs/videos if necessary._


dvdherron:
[Bug: Herman color preview styles are easily overridden by site custom styles.](https://trello.com/c/e6ymMLy8/110-bug-herman-color-preview-styles-are-easily-overridden-by-site-custom-styles)

dvdherron:
@mirisuzanne @stacyk 

I think what I have done here so far will prevent the font-size of Herman color swatches being overriden by custom sites. This selector is overriding padding, color, and background properties set in Herman though. 

bild

Ideas for the best approach to set values in Herman that won't be overriden by this? Or should this be adjusted on OddSite?

stacyk:
> 
> Ideas for the best approach to set values in Herman that won't be overriden by this? Or should this be adjusted on OddSite?

image

Is there anyway we can automatically add a class or attribute to these inline code snippets on OddSite? If not, maybe we can be a little more specific on OddSite and add a class (or use something like the h-entry class). I think this is more of an OddSite specificity issue.

mirisuzanne:
Trying to think through this from a structural perspective, because it feels like we might have a higher-level issue to solve.

- There are some places where we _only_ want Herman styles, and we need basically all Herman styles: e.g. most of the generated pages/documentation that is not in a nested iframe.
- Then there are some places where we _only_ want project styles, and we need basically all project styles: e.g. rendering sassdoc examples.

Those should both work ok with the current setup. The outer pages get only the full Herman stylesheet, and the example iframes get only the custom CSS set by the project. Is that right?

So then this issue is specifically about design-token previews (colors, fonts, sizes) which require:

- access to custom properties from the documented-project
- visualizations provided by Herman

And the issue is that both Herman and the documented-project are supplying more styles than necessary in those previews, right? We're getting the full custom CSS from the project (which can interfere with our previews), and maybe we're also passing in more Herman styles than we need for the previews? A few thoughts, which we could try some or all of:

- Split out Herman 'iframe styles' into individual stylesheets for each preview type, so there's less Herman code involved
- Ask projects for a second custom stylesheet, with only design tokens in it, so there's less project code interfering with previews
- Use cascade layers to ensure that Herman preview styles take precedence in selector conflicts. This doesn't solve the issue for properties we haven't set, but maybe using `all:revert` in a few places would do it?

What we really want here is 'scoping' to keep the the project styles from even accidentally selecting parts of our previews. First thought was shadow-DOM with custom components, but I'm not sure that would actually solve the problem?

Whatever we do here, shouldn't require making changes to OddSite styles in order to resolve the conflicts. It should be on Herman to make sure this all works right.

mirisuzanne:
@dvdherron @stacyk - I just chatted with @jgerigmeyer about this on our 1:1 call. We think it might be possible to auto-generate a version of the project CSS file that is stripped down to only include custom-properties. That would remove the issue of project styles interfering with preview styles. We'd still want to add these prefixes to Herman custom properties, to avoid naming conflicts in the tokens themselves.

jgerigmeyer:
> We think it might be possible to auto-generate a version of the project CSS file that is stripped down to only include custom-properties.

@dvdherron I can throw together a proof-of-concept for how this might work, if that'd be useful.

dvdherron:
> @dvdherron @stacyk - I just chatted with @jgerigmeyer about this on our 1:1 call. We think it might be possible to auto-generate a version of the project CSS file that is stripped down to only include custom-properties. That would remove the issue of project styles interfering with preview styles. We'd still want to add these prefixes to Herman custom properties, to avoid naming conflicts in the tokens themselves.

Ok, I'm guessing this would be all of the custom properties that we generate with Accoutrement (sizes/ratios, colors)? Or all custom properties in a project?

This probably means something like the [focus styling](https://www.oddbird.net/styleguide/config-focus) example wouldn't work or would have to be made in a different way?

lib/utils/getCustomProps.js
```diff
@@ -0,0 +1,30 @@
...
+    .findAll(
+      csstree.parse(css),
+      (node) =>
+        node.type === 'Rule' &&
+        node.prelude.children.some((sel) =>
+          sel.children.some((child) =>
+            ['html', 'root', 'body'].includes(child.name),
+          ),
+        ),
+    )
```
jgerigmeyer, line 17: Is this too specific? It will literally only find custom props declared in `html`, `:root`, or `body` rules (or some combination, e.g. `html, :root {...}`).

lib/utils/getCustomProps.js
```diff
@@ -0,0 +1,30 @@
...
+    .join('\n');
+  // Find all CSS custom properties declared in root nodes
+  const customProps = csstree.findAll(
+    csstree.parse(rootNodes),
+    (node) => node.type === 'Declaration' && node.property.startsWith('--'),
+  );
+  return `
+[data-herman-preview="colors"] {
+  ${customProps.map((decl) => csstree.generate(decl)).join(';\n  ')};
+}
```
jgerigmeyer, line 28: At this point I'm still throwing all the root rules together, then just extracting the custom properties and putting it in one new rule. Do we need to be concerned about conflicts between different blocks?

scss/initial/_root.scss
```diff
@@ -30,6 +31,7 @@
 /// @group style-typography
 :root {
   @include tools.font-family('sans');
+  @include tools.colors--(colors.$demo-colors);
```
jgerigmeyer, line 34: This maybe doesn't belong here? Not sure.

jgerigmeyer (commented):
@mirisuzanne @dvdherron @stacyk This is an initial proof-of-concept that seems to work for Herman itself, but I'm not sure about edge cases.

It also still needs tests, but I thought I'd wait to see if this is the direction we want to go first.

scss/initial/_root.scss
jgerigmeyer: @dvdherron I think the intent is to _not_ load them inside the iframe styles. The distinction is that anything included in `iframes.scss` will be made available within the iframe for _every project that uses Herman_. Anything that is specific to Herman's own documentation of itself (easily confused with Herman's role documenting other projects) should be in `main.scss`.

lib/utils/getCustomProps.js
mirisuzanne: I think it's a reasonable limitation for an automated approach, and we could also allow projects to pass in a custom design tokens file that they generate themselves if their general CSS doesn't follow this pattern.

lib/utils/getCustomProps.js
mirisuzanne: If we keep the restriction to three selectors:
- `:root` should override `html`
- `body` should override both `:root` and `html`
- if there are multiple from the same selector, last takes precedence

mirisuzanne (approved):

scss/iframes/_base.scss
```diff
@@ -1,3 +1,10 @@
+@use '~accoutrement/sass/tools';
```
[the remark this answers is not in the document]
dvdherron: Only in the previews for colors/fonts/sizes/etc @mirisuzanne

lib/utils/getCustomProps.js
jgerigmeyer: I'm tempted to say we don't try to enforce precedence, but we add an option for a user-supplied stylesheet as well.

stacyk (approved):
This all sounds like a good approach to move forward with.

lib/utils/getCustomProps.js
jgerigmeyer: Added in https://github.com/oddbird/sassdoc-theme-herman/commit/d5f6a2b4104c078414f2603781d0ce38c7467193

lib/utils/getCustomProps.js
jgerigmeyer: Note added in https://github.com/oddbird/sassdoc-theme-herman/commit/fb1b633ba7abab36db7a2ded060dd6116e72c866

scss/initial/_root.scss
jgerigmeyer: Moved to custom-props specific stylesheet in https://github.com/oddbird/sassdoc-theme-herman/commit/d5f6a2b4104c078414f2603781d0ce38c7467193

scss/config/_abstracts.scss
```diff
@@ -2,7 +2,7 @@
 @use 'sass:meta';
 @use '../utilities';
 @use '~accoutrement/sass/tools' with (
-  $color-var-prefix: ''
```
jgerigmeyer, line 5: I don't really understand how this works, but is this still the right place for it?

jgerigmeyer (approved):
@mirisuzanne @dvdherron I think this is ready for final review. I added a new `customPropertiesCSS` option if users want to use that instead (which Herman is currently doing).

scss/config/_abstracts.scss
dvdherron: I think ideally we would have this set up like on OddSite and TimeDesigner where we are only loading a configured Accoutrement in one place and then `@use`ing that file in all the places we are currently loading Accoutrement. 

We also have this Accoutrement configuration pattern at the top of `config/index` in other projects (Estimating Engine, Top Dog). Would that be the best place to relocate this in the meantime @mirisuzanne ? Or should we re-work how we have Accoutrement set up before this is merged/released?

dvdherron (commented):
can't approve since I made the PR, but this LGTM

scss/config/_abstracts.scss
mirisuzanne: @dvdherron I agree on both fronts:

- long term, we could update to match our usual pattern
- short term, we just need to make sure the configuration is the first use - so top of `config/index` makes sense for that

dvdherron: fixed in https://github.com/oddbird/sassdoc-theme-herman/pull/390/commits/b57cff56dc07d0d93726344df8cf80e6afeb16a1

jgerigmeyer (approved):
@mirisuzanne This looks good to me -- final review from you?

CONFIGURATION.md
```diff
@@ -345,15 +305,36 @@ as that output will be displayed in every single Sass example.
 
 [dart-sass-modules]: https://sass-lang.com/blog/the-module-system-is-launched
 
-### sass.outputStyle
+### sass.includes
```
mirisuzanne, line 308:
```suggestion
### sass.includes (deprecated)
```

I don't know, seems like we could make this clear in the label?

mirisuzanne (approved):
I haven't tried running it locally, but this all looks good to me.

PR #416, opened 2024-02-21
Support newer Sass `pkg` and compiler features.

jgerigmeyer:
- uses the shared resources compiler for `@example` annotations
- adds `NodePackageImporter` by default
- allows using sass-embedded for `@example` annotations
- uses sass-embedded internally throughout, except for the Sass tests (requires a change to True) -- local build time decrease from ~15sec to ~9sec

### DIFF
--- a/lib/annotations/example.js
+++ b/lib/annotations/example.js
@@ -81,6 +81,7 @@ module.exports = (env) => {
                 }
               }
             }
+            /* istanbul ignore else */
             if (sass && !sassCompiler) {
               try {
                 sassCompiler = sass.initAsyncCompiler();
--- a/lib/annotations/example.js
+++ b/lib/annotations/example.js
@@ -23,7 +23,7 @@ module.exports = (env) => {
     // eslint-disable-next-line global-require
     require('sassdoc/dist/annotation/annotations/example').default;
   const baseExample = baseExampleFn();
-  let sass;
+  let sass, sassCompiler;
   return {
     ...baseExample,
     resolve: (data) => {
@@ -61,13 +61,33 @@ module.exports = (env) => {
             exampleItem.rendered = undefined;
             /* istanbul ignore else */
             if (!sass) {
+              if (
+                env.herman?.sass?.implementation &&
+                typeof env.herman.sass.implementation !== 'string'
+              ) {
+                sass = env.herman.sass.implementation;
+              } else {
+                const sassPkg = env.herman?.sass?.implementation ?? 'sass';
+                try {
+                  // eslint-disable-next-line global-require
+                  sass = require(sassPkg);
+                } catch (err) /* istanbul ignore next */ {
+                  env.logger.warn(
+                    `Cannot find Dart Sass (\`${sassPkg}\`) dependency, ` +
+                      'which is required when using the `@example scss` annotation.',
+                  );
+                  env.logger.warn(err);
+                  return;
+                }
+              }
+            }
+            if (sass && !sassCompiler) {
               try {
-                // eslint-disable-next-line global-require
-                sass = require('sass');
+                sassCompiler = sass.initAsyncCompiler();
               } catch (err) /* istanbul ignore next */ {
                 env.logger.warn(
-                  'Cannot find Dart Sass (`sass`) dependency, which is ' +
-                    'required when using `@example scss` annotation.',
+                  'Cannot find Sass `initAsyncCompiler`,' +
+                    ' which is available since Dart Sass v1.70.0.',
                 );
                 env.logger.warn(err);
                 return;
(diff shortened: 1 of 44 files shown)


lib/annotations/example.js
```diff
@@ -61,28 +61,43 @@ module.exports = (env) => {
...
+                  return;
+                }
+              }
+            }
+            /* istanbul ignore else */
+            if (sass && !sassCompiler) {
               try {
-                // eslint-disable-next-line global-require
-                sass = require('sass');
+                sassCompiler = sass.initAsyncCompiler();
```
jamesnw, line 87: Option- I'm thinking it would be simpler and avoid a lot of subsequent `.then(compiler=>)` code if we could have `sassCompiler` be the compiler, and not the promise.

lib/annotations/example.js
```diff
@@ -61,28 +61,43 @@ module.exports = (env) => {
             exampleItem.rendered = undefined;
             /* istanbul ignore else */
             if (!sass) {
+              if (
```
jamesnw, line 64: I'm wondering of this new code might make sense factored out as a util instead of inline?

lib/annotations/example.js
```diff
@@ -121,9 +147,16 @@ module.exports = (env) => {
               sassPromises.push(promise);
             }
           }
-          const iframePromise = Promise.all(sassPromises).then(() =>
-            renderIframe(env, exampleItem, 'example'),
-          );
+          const iframePromise = Promise.all(sassPromises)
+            .then(() => {
```
jamesnw, line 151: Should this be `finally` instead of `then` so that the compiler is also disposed if there's an error?

jamesnw (approved):
Some minor nits, but overall looks good!

lib/annotations/example.js
jgerigmeyer: Moved in https://github.com/oddbird/sassdoc-theme-herman/pull/416/commits/f30775eaffcfa225aa51781ff93e81c45ea548ad

lib/annotations/example.js
jgerigmeyer: Hm... 🤔 I'm not excited about refactoring everything here to use async/await, and I'm not sure I see a clean way using the current Promise structure to guard against race conditions using your approach. I'm open to it if you have a proposal I'm missing, but otherwise since it's only used in two places total I'm also okay leaving it as-is.

jamesnw: That makes sense to me!

mirisuzanne (approved):
Again, trusting you all to test the functionality - lgtm

package.json
```diff
@@ -15,6 +15,11 @@
   "keywords": [
     "sassdoc-theme"
   ],
+  "main": "./index.js",
+  "exports": {
+    "sass": "./scss/_utilities.scss",
```
jamesnw, line 20: Should the docs at  https://www.oddbird.net/herman/docs/api_json-export be updated to recommend using `pkg:sassdoc-theme-herman`? It also strikes me as odd to have a utility be the main export, it could be available at `./utilities` or something, but that's also odd because then this doesn't have a default export.

jamesnw (approved):

package.json
jgerigmeyer: > Should the docs at https://www.oddbird.net/herman/docs/api_json-export be updated to recommend using `pkg:sassdoc-theme-herman`?

Hm, that's a good question. It's difficult to use `pkg:` in the actual compiled examples, because I don't think Herman can also have Herman installed in `node_modules`? 🤯 But at the very least we could add a comment documenting the recommended usage -- I'll add that.

> It also strikes me as odd to have a utility be the main export, it could be available at `./utilities` or something, but that's also odd because then this doesn't have a default export.

In this case `scss/_utilities.scss` is really the only file that users should ever be importing, so I think this makes sense? I guess we could rename it though.

jamesnw: > I don't think Herman can also have Herman installed

Possible, but that way lies danger :-D. I didn't do much digging on how the docs were built.

I don't think it's worth renaming

jgerigmeyer: > Possible, but that way lies danger :-D. I didn't do much digging on how the docs were built.

Yeah, I ran into this same issue in True when I tried to test using `pkg:sass-true` and realized that `sass-true` wasn't a valid installed package. 🤦

jgerigmeyer: Added in https://github.com/oddbird/sassdoc-theme-herman/pull/416/commits/81250ed5aad66c92db4a930b2aaceb2c1aa62b82

@mirisuzanne @jamesnw Do you think that's clear enough?

jamesnw: I was just thinking through if we should have added some method for self-referencing/internal pkg urls to Sass, but... nah.

PR #449, opened 2024-11-08
Upgrade deps and address Sass deprecation warnings.

jgerigmeyer:


scss/config/_colors.scss
```diff
@@ -62,17 +62,23 @@ $neutral-colors: (
   'light-gray': '#brand-blue'
     (
       'tint': 80%,
-      'desaturate': 80%,
+      'color.scale': (
```
jgerigmeyer, line 65: Should these be using `color.adjust` instead?

jgerigmeyer (commented):
@mirisuzanne @stacyk I don't really understand the changes I'm making to Sass, so I'd appreciate review from one or both of you! 🙏

scss/config/_colors.scss
stacyk: @jgerigmeyer in our case since we want to match the colors we have already established and tested, `color.adjust` will give us the same colors we had before.

jgerigmeyer: Fixed in https://github.com/oddbird/sassdoc-theme-herman/pull/449/commits/564f6638e44068715119475d462b944826336bcf

stacyk (approved):
This looks good to me. These changes match the desired output (some rounding of the hsl is the exception, but hex remains the same).

